PR #110, opened 2025-08-13
chore: update all packages and attempt to fix github action

rawkode:


zimbatm:
thanks! at some point I really need to write those update scripts myself instead of letting Claude do some half-baked job :sweat_smile:

rawkode:
Nah, let it keep going. Hell, I'm letting it build my own Nix native Direnv + Bazel tool and with enough guidance it's doing really well!

https://github.com/rawkode/cuenv

zimbatm:
nice. let's add cuenv to the list of related projects on the direnv.net website once it's in a good shape.

rawkode:
It's mostly just for fun and trying to solve the Nix problem of generic task orchestration, such as CD pipelines.

It's been really nice to explore as it reveals a lot of the subtle complexities a project like direnv has that aren't really obvious when you think about what the project does. 

So thank you

PR #135, opened 2025-08-16
qwen-code: 0.0.6 -> 0.0.7

lonerOrz:
Upgrade `qwen-code` from 0.0.6 to 0.0.7 and improve `update.sh`.

### DIFF
--- a/packages/qwen-code/package.nix
+++ b/packages/qwen-code/package.nix
@@ -1,45 +1,41 @@
 {
   lib,
+  nodejs_20,
   buildNpmPackage,
-  fetchFromGitHub,
+  fetchurl,
   fetchNpmDeps,
+  runCommand,
 }:
-
-buildNpmPackage (finalAttrs: {
+let
   pname = "qwen-code";
-  version = "0.0.6";
+  version = "0.0.7";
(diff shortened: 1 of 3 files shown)


packages/qwen-code/package.nix
```diff
@@ -1,53 +1,52 @@
...
+  src = runCommand "gemini-cli-src-with-lock" { } ''
+    mkdir -p $out
+    tar -xzf ${
+      fetchurl {
+        url = "https://registry.npmjs.org/@qwen-code/qwen-code/-/qwen-code-${version}.tgz";
+        hash = "${srcHash}";
+      }
+    } -C $out --strip-components=1
+    cp ${./package-lock.json} $out/package-lock.json
+  '';
```
zimbatm, line 24: It's preferable to build from source if we can. Was there a specific reason to switch?

zimbatm:
Fixed in #136. Thanks!

packages/qwen-code/package.nix
lonerOrz: I have encountered some cases where the package is not in the cache.

lonerOrz:
@zimbatm #136 is stuck.

zimbatm:
After taking a deeper look, your approach seems to be better

PR #175, opened 2025-08-21
fix: test formatter

zimbatm:
We were using the wrong prefix so the tests were not found by blueprint.


Mic92:
Had to disable mdsh but not too much of a problem because frequent updates will eventually fix our README anyhow: https://github.com/numtide/nix-ai-tools/pull/1034

PR #454, opened 2025-10-01
amp: pass in ripgrep and disable update check

keegancsmith:
Hello from the amp team. Amp will automatically download ripgrep if not found on PATH. So lets skip that and let nix manage ripgrep. Additionally we show a message if amp is out of date / auto-update fails. It will fail if amp is living in the nix store, so skip that step.


Mic92:
Thanks!

PR #661, opened 2025-10-21
feat(claudebox): add optional monitoring pane and improve usage docs

anstylian:
- Add `--no-monitor` option to run Claude without tmux monitoring pane
 - Update README with usage, options, and examples
 - Log commands to /tmp even when monitoring is disabled
 - Improve help output and argument parsing in claudebox.sh

### DIFF
--- a/packages/claudebox/README.md
+++ b/packages/claudebox/README.md
@@ -19,6 +19,7 @@ claudebox [OPTIONS]
 
 ### Options
 
+- `--no-monitor` - Skip tmux monitoring pane (run Claude directly in current terminal)
 - `--split-direction horizontal|vertical` - Set tmux split direction (default: `horizontal`)
 - `--no-tmux-config` - Don't load user tmux configuration (use default tmux settings)
 - `-h, --help` - Show help message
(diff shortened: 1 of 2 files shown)


zimbatm:
It looks good. The only thing I'm wondering is if it wouldn't make more sense to have it off by default, but to introduce an easy way to pop it open. Like add a tmux keyboard shortcut that opens a quake-like console. Is that possible?

anstylian:
> It looks good. The only thing I'm wondering is if it wouldn't make more sense to have it off by default, but to introduce an easy way to pop it open. Like add a tmux keyboard shortcut that opens a quake-like console. Is that possible?

I didn’t want to change the default behavior, so I kept it as is (though I’m open to it).

As far as I know, there’s no simple way to do that with plain `tmux`, but if there is, please let me know! That said, the user now has a couple of options:
1. They can zoom into the panel where Claude is running to focus on it and hide the logs when needed.
1. This PR moves the log file to `/tmp`, which is accessible to everyone. If `command-viewer` is installed as a package, the user can run it in a separate terminal to monitor Claude’s commands.

zimbatm:
Alright. Can you rebase the PR and also change the ENABLE_MONITOR to enable_monitor? Then it's good to go.

zimbatm:
LGTM, thanks!

PR #662, opened 2025-10-21
feat(claudebox): add tmux split direction and config options

anstylian:
- Add `--split-direction horizontal|vertical` to choose tmux pane layout
- Add `--no-tmux-config` to ignore user tmux configuration
- Update README with usage, options, and layout details
- Refactor script to support custom tmux config and split direction

### DIFF
--- a/packages/claudebox/claudebox.sh
+++ b/packages/claudebox/claudebox.sh
@@ -5,8 +5,19 @@ set -euo pipefail
 # Shows all commands Claude executes in a tmux split pane
 
 # Default options
-SPLIT_DIRECTION=horizontal
-LOAD_TMUX_CONFIG=true
+read -r rows cols < <(stty size 2>/dev/null || echo "24 80")
+# If the terminal is very wide (at least 3 times as wide as it is tall),
+# use a horizontal split (top/bottom panes).
+# Note: In tmux, "horizontal" refers to the dividing line being horizontal,
+# which produces stacked panes.
+if ((cols >= rows * 3)); then
+  split_direction="horizontal"
+else
+  # Otherwise, use a vertical split (side-by-side panes).
+  # Here, the dividing line is vertical, giving more height to each pane.
+  split_direction="vertical"
+fi
+load_tmux_config=true
 
 # Parse command-line arguments
 show_help() {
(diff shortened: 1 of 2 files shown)


packages/claudebox/claudebox.sh
```diff
@@ -4,6 +4,53 @@ set -euo pipefail
 # claudebox - Run Claude Code in YOLO mode with transparency
 # Shows all commands Claude executes in a tmux split pane
 
+# Default options
+SPLIT_DIRECTION=horizontal
+LOAD_TMUX_CONFIG=true
```
zimbatm: I reserve UPPER_CASE vars to environment variables. So use snake_case instead.

packages/claudebox/claudebox.sh
```diff
@@ -77,6 +127,16 @@ bwrap_args=(
...
 )
 
+# Mount tmux configuration (support both traditional and XDG locations)
+if [[ $LOAD_TMUX_CONFIG == "true" ]]; then
+  if [[ -f "${HOME}/.tmux.conf" ]]; then
+    bwrap_args+=(--ro-bind "${HOME}/.tmux.conf" "$HOME/.tmux.conf")
+  fi
+  if [[ -d "${HOME}/.config/tmux" ]]; then
+    bwrap_args+=(--ro-bind "${HOME}/.config/tmux" "$HOME/.config/tmux")
+  fi
```
zimbatm, line 152: Don't forget the XDG variant too:

```
FILES
       ~/.tmux.conf
       $XDG_CONFIG_HOME/tmux/tmux.conf
       ~/.config/tmux/tmux.conf
```

packages/claudebox/claudebox.sh
```diff
@@ -4,6 +4,53 @@ set -euo pipefail
 # claudebox - Run Claude Code in YOLO mode with transparency
 # Shows all commands Claude executes in a tmux split pane
 
+# Default options
+SPLIT_DIRECTION=horizontal
```
zimbatm: It would be nice if the default changed depending on the terminal dimensions. Not a blocker.

zimbatm (commented):
looks good overall, just a few comments

packages/claudebox/claudebox.sh
anstylian: I’ve implemented your suggestion. Please let me know if you have an alternative heuristic in mind.

zimbatm:
thanks!

PR #709, opened 2025-10-25
 ECA - Editor Code Assistant 

zrubing:
add x86_64_linux eca package

### DIFF
--- a/packages/eca/package.nix
+++ b/packages/eca/package.nix
@@ -1,7 +1,7 @@
 { pkgs }:
 
 let
-  version = "0.73.4";
+  version = "0.73.5";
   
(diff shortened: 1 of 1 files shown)


packages/eca/package.nix
```diff
@@ -0,0 +1,90 @@
...
+      stripRoot = false;
+    };
+
+    dontUnpack = true;
+    dontBuild = true;
+
+    installPhase = ''
+      runHook preInstall
+      mkdir -p $out/bin
+      cp $src/eca $out/bin/eca
```
Mic92, line 22: is this a static binary?

zrubing: yes

PR #917, opened 2025-11-11
claude-desktop: add desktop file & uri handler

baduhai:
This pr adds a desktop file for the claude-desktop package. The desktop package alse makes the program handle the claude:// uri.

### DIFF
--- a/packages/claude-desktop/package.nix
+++ b/packages/claude-desktop/package.nix
@@ -3,11 +3,14 @@
   stdenv,
   fetchurl,
   makeWrapper,
+  makeDesktopItem,
+  copyDesktopItems,
   p7zip,
   unzip,
   electron,
   nodejs,
   asar,
+  graphicsmagick,
 }:
 
 let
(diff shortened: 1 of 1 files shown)


Mic92:
Interesting bug you triggered in nix:

terminate called after throwing an instance of 'boost::wrapexcept<boost::io::too_few_args>'

baduhai:
> Interesting bug you triggered in nix:

The true power of vibe coding.

packages/claude-desktop/package.nix
```diff
@@ -167,6 +189,10 @@ stdenv.mkDerivation rec {
       --set DISABLE_AUTOUPDATER 1 \
       --set NODE_ENV production
 
+    # Install icon
+    mkdir -p $out/share/icons/hicolor/scalable/apps
+    cp ${icon} $out/share/icons/hicolor/scalable/apps/claude-desktop.svg
```
baduhai: After manual review, this line (194) is the culprit. Perhaps the icon path has a character boost really doesn't like when interpolating nix to bash?

baduhai:
Looks like the checks are now passing, the latest commit uses the icon available in the installer's exe, which is better than fetching an svg from Wikimedia anyway.

packages/claude-desktop/package.nix
```diff
@@ -60,37 +63,71 @@ stdenv.mkDerivation rec {
...
+  ];
+
   unpackPhase = ''
     runHook preUnpack
 
     # Extract the Windows installer - use -y to auto-overwrite
     7z x -y $src -o./extracted
 
+    # Extract and convert icon to multiple PNG sizes
+    if [ -f ./extracted/setupIcon.ico ]; then
```
Mic92: @copilot we don't need this code here and than we loop over icons in the installPhase again. Just do the conversion in install phase in one loop. Also use graphicsmagic instead of imagemagic.

PR #926, opened 2025-11-12
opencode: 1.0.55 -> 1.0.62

delafthi:
- wrap binary with fzf and ripgrep runtime dependencies
- replace `skip-npm-pack.patch` with `skip-bun-install.patch` due to
upstream changes of the build script
- install all platform versions with `bun install` to make the
`node_modules` hash platform independent
- use isolated installs like upstream


delafthi:
I might have found a way to remove platform dependent hashes -> https://github.com/NixOS/nixpkgs/pull/460710. However, there I'm currently struggling with Bun v1.3.2.

Mic92:
```
       > building opencode-linux-arm64-musl
       > error: Failed to download 'bun-linux-aarch64-musl-v1.3.1': ConnectionRefused
       > 
```

Could we even skip this target?

Mic92:
We also don't need to build the baseline package:

```
       > building opencode-linux-x64-baseline
       > error: Failed to download 'bun-linux-x64-baseline-v1.3.1': ConnectionRefused
       > 
```

We can just patch it to build the `opencode-linux-x64` target.

delafthi:
@Mic92 thanks, finally I got what the issue is. I previously thought it's just a networking problem. 
I've patched the build script to skip those targets and updated the derivation to remove platform-specific hashes. I've also adapted `update.sh` accordingly, though the GitHub workflows still need to be updated.

Mic92:
I also tried to use the bundle instead of compile. This works on macOS fine but now I seem to trigger a bug on Linux: https://github.com/numtide/nix-ai-tools/pull/942

Mic92:
there is also a merge conflict.

delafthi:
> there is also a merge conflict.

Resolved

delafthi:
> I also tried to use the bundle instead of compile. This works on macOS fine but now I seem to trigger a bug on Linux: #942

🙈 I hoped that bundling would help to bridge the time until the upcoming build issues with bun v1.3.2 are solved (https://github.com/oven-sh/bun/issues/24645).

delafthi:
Sorry for the issues. I've cleaned up the PR - it should be good to go now.

packages/opencode/update.sh
```diff
@@ -33,78 +36,60 @@ echo "Current version: $current_version"
...
-  export NIX_PATH=nixpkgs=flake:nixpkgs
-  hash_output=$(timeout 60 nix-build -E "with import <nixpkgs> {}; fetchzip { url = \"${url}\"; sha256 = \"\"; }" 2>&1 || true)
-  new_hash=$(echo "$hash_output" | grep "got:" | awk '{print $2}')
-
-  if [ -z "$new_hash" ]; then
-    echo "    ERROR: Failed to calculate hash for $nix_system"
-    echo "    Output: $hash_output"
-    continue
-  fi
+echo "Calculating source hash..."
```
Mic92, line 52: If we now only have a single hash we should also remove the custom github action, I built for opencode, we can than use the generic one that only runs on linux.

Mic92: If you remove that, I can merge this.

PR #942, opened 2025-11-13
 opencode: simplify build by bundling instead of compiling 

Mic92:


Mic92:
I give up on this approach.

PR #1111, opened 2025-11-26
refactor(claude-code): switch from npm build to binary download

ryoppippi:
Replace buildNpmPackage with pre-built binaries from Anthropic's distribution server. 

according to install page, they recommend binary version

https://code.claude.com/docs/en/setup#native-installation-options
> The native installation is the recommended method and offers several benefits:
One self-contained executable
No Node.js dependency
Improved auto-updater stability
If you have an existing installation of Claude Code, use claude install to migrate to the native binary installation.

i created a similar overlay, but i'd love to migrate it in this repp
https://github.com/ryoppippi/claude-code-overlay


ryoppippi:
I'm new at nix, so if my PR needs refactoring let me know. thanks

Mic92:
I am a bit worried how they package the binary. Is this using bun? I had issues with patchelf on arm64-linux before because bun uses some uncommon linker tricks

Mic92:
You can ask your llm to use `strings` to figure this out. If this doesn't explain it, `objdump` might.

ryoppippi:
yeah it is compiled with bun, same as coderabbit. i'll fix it if it failed

ryoppippi:
I asked Claude about it (as an OSS maintainer, I know this is rude, and I’m very sorry.

```
  I investigated the binary and confirmed it's built with Bun (found many @oven/bun-* package references via strings).

  You're right to be concerned - there are https://unix.stackexchange.com/questions/522822/different-methods-to-run-a-non-nixos-executable-on-nixos that
  bundle JS code as appended data to ELF executables. However:

  1. CI checks are passing - both buildbot/nix-build and nix-eval succeeded, which should include aarch64-linux builds
  2. coderabbit-cli uses the same approach - it's also a Bun binary with autoPatchelfHook and dontStrip = true, and it's been working
  3. The official Bun package in nixpkgs also uses autoPatchelfHook successfully

  One thing I noticed: the buildInputs includes zlib and stdenv.cc.cc.lib, but:
  - nixpkgs' Bun package only specifies openssl
  - coderabbit-cli has no additional buildInputs

  These might be unnecessary. I can test on aarch64-linux to verify which libraries are actually needed, and remove any that aren't required.
```

Mic92:
Most of these points actually don't matter all that much when using the nixpkgs:

> One self-contained executable
> No Node.js dependency
> Improved auto-updater stability
> If you have an existing installation of Claude Code, use claude install to migrate to the native binary installation.

Users might already have nodejs somewhere and in this way we actually save storage because we don't also have a bun runtime somewhere and the rest here also mostly doesn't matter since we package with nix. If we don't have a concrete reason to switch, I would like stick to the npm package, based on difficulties I had with opencode's bun binaries.

Thanks for your work.

ryoppippi:
Ok thanks!

PR #1160, opened 2025-11-30
fix(ccusage): replace bun to nodejs

ryoppippi:
based on this feedback, use nodejs instead of bun
https://github.com/numtide/nix-ai-tools/pull/1111

### DIFF
--- a/packages/ccusage-codex/package.nix
+++ b/packages/ccusage-codex/package.nix
@@ -2,7 +2,7 @@
   lib,
   stdenv,
   fetchzip,
-  bun,
+  nodejs,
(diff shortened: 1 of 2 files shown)


Mic92:
For that type of usage bun, might be fine actually. It's this type of usage that was problematic on macOS: https://bun.com/docs/bundler/executables

and for aarch64-linux it was patchelf that caused problems for converting existing packaged bun apps. However if you prefer nodejs, than we can also have this. Your call.

ryoppippi:
@Mic92 i see, if so i'll close it! thanks

PR #1256, opened 2025-12-06
claude-plugins: init at 0.2.0

YPares:
Added [`claude-plugins`](https://github.com/Kamalnrf/claude-plugins), which provides a more convenient CLI to install activate/deactivate plugins for Claude Code (without having to pre install & register marketplaces).
Also provides easier plugins discovery via https://claude-plugins.dev/

### DIFF
--- a/packages/claude-plugins/package.nix
+++ b/packages/claude-plugins/package.nix
@@ -35,7 +35,7 @@ stdenv.mkDerivation rec {
     changelog = "https://github.com/Kamalnrf/claude-plugins/releases";
     license = licenses.mit;
     sourceProvenance = with lib.sourceTypes; [ binaryBytecode ];
-    maintainers = with maintainers; [ ];
+    maintainers = with pkgs.maintainers; [ ypares ];
     mainProgram = "claude-plugins";
     platforms = platforms.all;
   };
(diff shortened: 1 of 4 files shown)


packages/claude-plugins/package.nix
```diff
@@ -0,0 +1,42 @@
...
+    runHook postInstall
+  '';
+
+  meta = with lib; {
+    description = "CLI tool for managing Claude Code plugins";
+    homepage = "https://github.com/Kamalnrf/claude-plugins";
+    changelog = "https://github.com/Kamalnrf/claude-plugins/releases";
+    license = licenses.mit;
+    sourceProvenance = with lib.sourceTypes; [ binaryBytecode ];
+    maintainers = with maintainers; [ ];
```
Mic92: Can you add yourself as maintainer here?

Mic92: if you are not a nixpkgs maintainer: https://github.com/numtide/llm-agents.nix/blob/main/lib/default.nix

YPares: Yep :) Done!

YPares:
@Mic92 I also added me as a maintainer of openskills & ccstatusline, as I did suggest adding them and they had none.

PR #1258, opened 2025-12-07
claude-code: fix build on macOS with `sandbox = true`

Enzime:


Enzime:
Seems like `.#claude-code.npmDeps` just hangs when built with sandbox now, I'll need to investigate further

Mic92:
Should we mark this as sandbox relaxed in the meantime?

Mic92:
PR seems to be stalled.

Enzime:
I'm in the process of investigating this further and I think it's an issue with the sandbox profile in Nix, after a while it eventually fails with `unable to resolve DNS hostname`

PR #1328, opened 2025-12-11
Link to new binary cache 

Mic92:


jonhoo:
Since this changes the server and public key for the binary cache, could you give a bit more context on the change? The empty description just makes me slightly nervous given the rise of supply chain risks.

Issue #399, opened 2025-09-25
opencode segfault nixos arm


u011:
**Describe the bug**

install exactly as readme, getting -> Job 1, 'opencode' terminated by signal SIGSEGV (Address boundary error)

**To Reproduce**

flake as in docs nixos 25 on arm in vmware on mac. gemini works, opencode not

---
 
Claude managed to run the binary by direct download like this:

Enabled nix-ld:
       54 +    # Enable nix-ld for running unpatched binaries
       55 +    programs.nix-ld = {
       56 +      enable = true;
       57 +      libraries = with pkgs; [
       58 +        stdenv.cc.cc.lib
       59 +        glibc
       60 +        zlib
       61 +        libgcc
       62 +      ];
       63 +    };
 
downloaded and run with curl -L https://github.com/sst/opencode/releases/download/v0.11.2/opencode-linux-arm64.zip -o  /tmp/opencode.zip && cd /tmp && unzip -o opencode.zip && ./opencod

Its explanation and fixes:

Why OpenCode crashed: The nix-ai-tools package uses autoPatchelfHook which incorrectly patches the ARM64 binary,
  causing segfaults in the dynamic linker. The solution was to create a custom package with manual LD_LIBRARY_PATH
  wrapping.

Would be nice if you could make it work without workarounds

Mic92:
LD_LIBRARY_PATH wouldn't help here on systems without nix-ld.

Mic92:
Switching to building from source might be an option, https://github.com/NixOS/nixpkgs/blob/master/pkgs/by-name/op/opencode/package.nix

I am currently don't need aarch64-linux, so feel free to take this on.

Mic92:
Fixed https://github.com/numtide/nix-ai-tools/pull/420

PR #1424, opened 2025-12-18
claude-code-native: init at 2.0.65

allouis:
## Summary

Adds `claude-code-native` package that installs Claude Code from the official native binary distribution rather than npm.

### Why native?

The [official documentation](https://code.claude.com/docs/en/setup#native-install-recommended) now recommends native installation over npm. [Some upcoming features will be native-only](https://x.com/dmwlff/status/2001030789065240601).

Note: The native release channel seems to trail the npm version slightly (currently 2.0.65 vs 2.0.71).

### Tested

- Built and ran on aarch64-darwin
- Built and ran on x86_64-linux
- Verified `update.py` fetches latest version and updates hashes correctly

---

### Questions

- Should this eventually replace `claude-code` entirely, given that native is now the recommended installation method?
- Not sure how you'll feel about the update script, I've fetched the install script each time and resolved the bucket URLs, just in case the bucket URLs end up changing - it feels a bit flakey, do you have better ideas for this?

### DIFF
--- a/packages/claude-code-native/update.py
+++ b/packages/claude-code-native/update.py
@@ -1,0 +1,61 @@
+#!/usr/bin/env nix
+#! nix shell --inputs-from .# nixpkgs#python3 --command python3
+
+"""Update script for claude-native package."""
+
+import sys
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent / "scripts"))
+
+from updater import (
+    calculate_platform_hashes,
+    fetch_text,
+    load_hashes,
+    save_hashes,
+    should_update,
+)
+
+HASHES_FILE = Path(__file__).parent / "hashes.json"
+INSTALL_SCRIPT_URL = "https://claude.ai/install.sh"
+
+PLATFORMS = {
+    "x86_64-linux": "linux-x64",
+    "aarch64-linux": "linux-arm64",
+    "x86_64-darwin": "darwin-x64",
+    "aarch64-darwin": "darwin-arm64",
+}
+
+
+def get_bucket_url() -> str:
+    """Discover bucket URL from the official install script."""
+    script = fetch_text(INSTALL_SCRIPT_URL)
+    for line in script.splitlines():
+        if line.startswith("GCS_BUCKET="):
+            return line.split("=", 1)[1].strip('"')
+    msg = "Could not find GCS_BUCKET in install script"
+    raise ValueError(msg)
+
+
+def main() -> None:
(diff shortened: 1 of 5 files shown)


Mic92:
We had some issues with patchelf + aarch64-linux for pre-compiled bun binaries. https://github.com/numtide/llm-agents.nix/issues/399

Are there any other functional advantages of using bun here besides a recommendation?
Bun's binary installation does some nifty elf linking don't seem to work well with patchelf.

allouis:
Right now the functional advantage is syntax highlighting in the code diffs proposed by Claude - but according to this tweet https://x.com/dmwlff/status/2001030789065240601 there will be future features that will be native only too

Mic92:
Okay. I have to test this a bit on Linux. Maybe there is also a way to avoid patchelf all-together.

allouis:
Alright - thanks for looking into this 🙏

packages/claude-code-native/update.py
```diff
@@ -0,0 +1,61 @@
...
+    "x86_64-darwin": "darwin-x64",
+    "aarch64-darwin": "darwin-arm64",
+}
+
+
+def get_bucket_url() -> str:
+    """Discover bucket URL from the official install script."""
+    script = fetch_text(INSTALL_SCRIPT_URL)
+    for line in script.splitlines():
+        if line.startswith("GCS_BUCKET="):
```
Mic92, line 34: Relying on the structure of a script looks a bit more brittle compared to what @ryoppippi has figured out in: https://github.com/ryoppippi/nix-ai-tools/blob/bf4dee2bb6df42ee4cfd7efb2622cba047115ef9/packages/claude-code/update.py#L44

Mic92: There also seem to be a manifest that already contain our checksums: https://github.com/ryoppippi/nix-ai-tools/blob/bf4dee2bb6df42ee4cfd7efb2622cba047115ef9/packages/claude-code/update.py#L58

Mic92: Here is the original pr: https://github.com/numtide/llm-agents.nix/pull/1111

Mic92:
Thanks for the research in this pull request. I recently introduced `wrapBuddy`, which shouldn't cause the crashes we are seeing on aarch64-linux. I added the changes here: https://github.com/numtide/llm-agents.nix/pull/1515 to replace the claude-code package.
and used the manifest file for updating that https://github.com/numtide/llm-agents.nix/pull/1111 introduced.

Mic92:
Done

PR #1515, opened 2025-12-23
claude-code: switch from npm build to native binary distribution

Mic92:
This replaces the buildNpmPackage approach with pre-built binaries from Anthropic's official distribution server, as recommended in their docs.

The update script now fetches version info from the "latest" channel and extracts checksums directly from manifest.json, eliminating the need for package-lock.json generation.

Uses wrapBuddy instead of autoPatchelfHook for Linux binary patching.

Based on the approach from PR #1111 by @ryoppippi.


Mic92:
@allouis @ryoppippi can you test?

PR #1684, opened 2026-01-03
pi: init at 0.34.2

aos:
This adds https://github.com/badlogic/pi-mono/tree/main/packages/coding-agent The agent is part of a monorepo. See also: https://shittycodingagent.ai/

I opted for using the binary package name `pi` rather than `pi-coding-agent` as is the NPM package name.

Edit: updated to `0.34.2` since that was released a bit ago.

### DIFF
--- a/README.md
+++ b/README.md
@@ -371,6 +371,16 @@ Nix packages for AI coding agents and development tools. Automatically updated d
 
 </details>
 <details>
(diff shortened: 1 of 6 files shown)


aos:
Welp... looks like `v0.33.0` broke aarch64 builds. Filing an issue upstream!

aos:
I switched it to using `buildNpmPackage` instead as that seems to work just fine. Looks like it's some sort of bun issue. See my comment at https://github.com/badlogic/pi-mono/issues/438#issuecomment-3708505669

PR #1694, opened 2026-01-04
fetchNpmDeps: rename cacheVersion to fetcherVersion

Mic92:
Since we now write .fetcher-version file in the cache output, all hashes for fetcherVersion 2 have changed.


Mic92:
Rebased to match our nixpkgs pull request.

PR #1754, opened 2026-01-07
prefetch-npm-deps: normalize npm api responses.

Mic92:
See https://github.com/numtide/llm-agents.nix/commit/86df9d9fb28ccb6d290c6f1c607a249543d96ef1

### DIFF
--- a/packages/amp/hashes.json
+++ b/packages/amp/hashes.json
@@ -1,5 +1,5 @@
 {
   "version": "0.0.1767787299-gb00fdc",
   "sourceHash": "sha256-7MxrPiDzpt9g89Y6xyZ8w20afFx3ddfdf8MMuyt4Emc=",
-  "npmDepsHash": "sha256-IfNPGD1YDCr/6iQ14qlaAtpQzY4cKQUMzEPy4P+H4bM="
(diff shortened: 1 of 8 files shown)


clefru:
Are the volatile npm API failures always such that sometimes some fields are empty? Would it be possible to retry the request after 1, 2, 4, 8 seconds (exponential backoff) and build the union of all responses?

Mic92:
> Are the volatile npm API failures always such that sometimes some fields are empty? Would it be possible to retry the request after 1, 2, 4, 8 seconds (exponential backoff) and build the union of all responses?

The commit message is more up-to-date than my PR title (which is a bit misleading):

https://github.com/numtide/llm-agents.nix/pull/1754/changes/86df9d9fb28ccb6d290c6f1c607a249543d96ef1

Mic92:
It's more volatile in a sense that data for existing package versions seems to change, so I am excluding fields.

clefru:
Oh ic, thanks for the explanation. The motivation bug report just made it look like that sometimes fields are empty.

Thanks for fixing this!

Mic92:
> Oh ic, thanks for the explanation. The motivation bug report just made it look like that sometimes fields are empty.
> 
> Thanks for fixing this!

Let's see if that works. Time will tell. I removed the attack surface now quiet a bit.

PR #1802, opened 2026-01-09
pi: add makeWrapper to fix env variable in PATH

aos:
Looks like wrapProgram somehow still works to add `fd` and `ripgrep` even when `makeWrapper` is not included, but it fails to update the path. I needed to add it so that it also updates the path to add `PI_SKIP_VERSION_CHECK`

### DIFF
--- a/packages/pi/package.nix
+++ b/packages/pi/package.nix
@@ -4,6 +4,7 @@
   fetchurl,
   fetchNpmDepsWithPackuments,
   npmConfigHook,
+  makeWrapper,
   fd,
   ripgrep,
   runCommand,
@@ -40,6 +41,8 @@ buildNpmPackage {
     cacheVersion = 2;
   };
   makeCacheWritable = true;
+
+  nativeBuildInputs = [ makeWrapper ];
 
   # The package from npm is already built
   dontNpmBuild = true;


packages/pi/package.nix
```diff
@@ -41,6 +42,8 @@ buildNpmPackage {
   };
   makeCacheWritable = true;
 
+  nativeBuildInputs = [ makeWrapper ];
```
Mic92, line 45: Not sure this is actually need works fine without:

```
~/git/llm-agents.nix main
% nix build .#pi

~/git/llm-agents.nix main
% cat ./result/bin/pi
─────┬────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
     │ File: ./result/bin/pi
─────┼────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
   1 │ #! /nix/store/yi3x6xpilh0bq0bgb61imy6gpv8b1la4-bash-5.3p3/bin/bash -e
   2 │ PATH=${PATH:+':'$PATH':'}
   3 │ PATH=${PATH/':''/nix/store/zydgmksf6gf4q95k43qym4nzf0w4881a-ripgrep-15.1.0/bin'':'/':'}
   4 │ PATH='/nix/store/zydgmksf6gf4q95k43qym4nzf0w4881a-ripgrep-15.1.0/bin'$PATH
   5 │ PATH=${PATH#':'}
   6 │ PATH=${PATH%':'}
   7 │ export PATH
   8 │ PATH=${PATH:+':'$PATH':'}
   9 │ PATH=${PATH/':''/nix/store/zc0027xm116rwk8xd4hxs24zx4kn6lqz-fd-10.3.0/bin'':'/':'}
  10 │ PATH='/nix/store/zc0027xm116rwk8xd4hxs24zx4kn6lqz-fd-10.3.0/bin'$PATH
  11 │ PATH=${PATH#':'}
  12 │ PATH=${PATH%':'}
  13 │ export PATH
  14 │ export PI_SKIP_VERSION_CHECK='1'
  15 │ exec -a "$0" "/nix/store/20dz3zf5hnllykjs14ba9h2cdd44264h-pi-0.42.0/bin/.pi-wrapped"  "$@"
─────┴────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
```

Mic92: the package would not build otherwise if wrapProgram was missing.

Mic92:
Closing this for now.

aos:
Okay yeah - thanks. Apparently I had the version _right before_ the environment variable was added. Was confused because it was using `0.42.0` and I think that update landed one hour before :facepalm:

PR #1837, opened 2026-01-12
chainlink: init at 1.4

Chickensoupwithrice:
## Summary
- Add chainlink package - a simple, lean issue tracker CLI designed for AI-assisted development
## Details
- **Homepage**: https://github.com/dollspace-gay/chainlink
- **License**: MIT
- **Category**: Workflow & Project Management
### Implementation notes
- Uses \`cargoRoot\` and \`buildAndTestSubdir\` since the Rust crate is in a subdirectory
- Patches \`Cargo.toml\` version during build (upstream uses \`0.1.0\` while releases are \`chainlink-X.Y\`)
- Custom \`nix-update-args\` with version regex for non-standard tag format (\`chainlink-1.4\` instead of \`v1.4\`)
- Version check enabled and tested
### Testing

```
$ nix build .#chainlink
$ ./result/bin/chainlink --version
chainlink 1.4.0
```
nix-update tested successfully (downgraded to 1.3, ran nix-update, correctly updated to 1.4).

### DIFF
--- a/packages/chainlink/package.nix
+++ b/packages/chainlink/package.nix
@@ -1,5 +1,6 @@
 {
   lib,
+  flake,
   fetchFromGitHub,
   rustPlatform,
   versionCheckHook,
@@ -41,6 +42,7 @@ rustPlatform.buildRustPackage rec {
     homepage = "https://github.com/dollspace-gay/chainlink";
     license = licenses.mit;
     sourceProvenance = with sourceTypes; [ fromSource ];
+    maintainers = with flake.lib.maintainers; [ Chickensoupwithrice ];
     mainProgram = "chainlink";
     platforms = platforms.all;
   };
--- a/lib/default.nix
+++ b/lib/default.nix
@@ -12,6 +12,11 @@ inputs.nixpkgs.lib.extend (
         githubId = 1377233;
         name = "Yves Parès";
       };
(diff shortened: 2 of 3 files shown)


packages/chainlink/package.nix
```diff
@@ -0,0 +1,47 @@
...
+  doInstallCheck = true;
+  nativeInstallCheckInputs = [ versionCheckHook ];
+
+  passthru.category = "Workflow & Project Management";
+
+  meta = with lib; {
+    description = "Simple, lean issue tracker CLI designed for AI-assisted development";
+    homepage = "https://github.com/dollspace-gay/chainlink";
+    license = licenses.mit;
+    sourceProvenance = with sourceTypes; [ fromSource ];
```
Mic92, line 44: Do you want to maintain this package @Chickensoupwithrice
I currently don't want to, we have similar packages already and this one is just 2 weeks old.

Chickensoupwithrice: I'm happy to maintain it, will remove it if I no longer use it.

Mic92: Okay. Please add yourself as a maintainer as described here: https://github.com/numtide/llm-agents.nix/blob/main/AGENTS.md#custom-maintainers

PR #1966, opened 2026-01-17
Add overlay support with llm-agents namespace

Xantibody:
## Summary

  - Add `overlays.default` output to flake
  - Packages available under `pkgs.llm-agents.*` namespace (e.g., `pkgs.llm-agents.claude-code`)
  - Utility packages excluded from overlay (formatter, flake-inputs, etc.)
  - Document overlay usage in README

  ## Implementation

  - `overlays/default.nix`: Overlay definition using `final.stdenv.hostPlatform.system`
  - `flake.nix`: Export overlay from blueprint outputs
  - `README.md`: Add "Using Overlay" section

  ## Test plan

  - [x] `nix flake show` shows `overlays.default`
  - [x] `nix eval` confirms overlay provides `llm-agents` namespace with packages
  - [x] Overlay works in nix-darwin configuration with `pkgs.llm-agents.claude-code`
  - [x] `nix fmt` passes
  - [x] `nix flake check` (currently running)

## Motivation & Note
Having experienced the convenience of overlays in other modules, I’ve implemented this to improve the developer experience and keep configurations cleaner. If there’s a specific design reason why an overlay hasn't been provided yet, I’m happy to discuss or close this PR according to your project goals.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -25,20 +25,11 @@
         nixpkgs.config.allowUnfree = true;
       };
 
-      # Utility packages to exclude from the overlay
-      excludedPackages = [
-        "formatter"
-        "flake-inputs"
-        "versionCheckHomeHook"
-        "darwinOpenptyHook"
-        "wrapBuddy"
-      ];
     in
     blueprintOutputs
     // {
       overlays.default = import ./overlays {
         packages = blueprintOutputs.packages;
-        inherit excludedPackages;
       };
     };
 }
--- a/overlays/default.nix
+++ b/overlays/default.nix
@@ -1,8 +1,6 @@
 {
   packages,
-  excludedPackages,
 }:
 final: _prev: {
-  llm-agents = builtins.removeAttrs (packages.${final.stdenv.hostPlatform.system} or { }
-  ) excludedPackages;
+  llm-agents = packages.${final.stdenv.hostPlatform.system} or { };
 }


Mic92:
> Having experienced the convenience of overlays in other modules, I’ve implemented this to improve the developer experience and keep configurations cleaner. If there’s a specific design reason why an overlay hasn't been provided yet, I’m happy to discuss or close this PR according to your project goals.

I guess we haven't added overlays because me and zimbatm, while often used, we don't think its more readable/cleaner to use overlays. It makes it hard to track in bigger projects where packages are coming from because everything is merged into one big set. Also it has a measurable performance impact to add overlays as it adds another fixed point computation when accessing anything from pkgs (nix does not have lazy data structures so any overlay forces also a copy of your big package attrset). Overlays seems to be commonly used to adding packages to a NixOS configuration, but usually using `specialArgs` is a better way to make it obvious where packages are coming from.

README.md
```diff
@@ -563,6 +563,34 @@ Add to your system configuration:
 }
 ```
 
+### Using Overlay
```
Mic92, line 566: If you add this, can you add `(recommend)` to Using flakes.

flake.nix
```diff
@@ -19,8 +19,26 @@
...
+        "formatter"
+        "flake-inputs"
+        "versionCheckHomeHook"
+        "darwinOpenptyHook"
+        "wrapBuddy"
+      ];
+    in
+    blueprintOutputs
+    // {
+      overlays.default = import ./overlays {
```
Mic92, line 31: @zimbatm is there a different way to add overlays in blueprint?

flake.nix
```diff
@@ -19,8 +19,26 @@
...
+      };
+
+      # Utility packages to exclude from the overlay
+      excludedPackages = [
+        "formatter"
+        "flake-inputs"
+        "versionCheckHomeHook"
+        "darwinOpenptyHook"
+        "wrapBuddy"
+      ];
```
Mic92: Since everything is namespaced under `llm-agents`, I rather not want to have to maintain this list for a feature that I commonly not use.

```suggestion
```

Xantibody: Fixed, thanks!

README.md
Xantibody: Thank you for the suggestion! That makes total sense — added "(Recommended)" to the flakes section

Xantibody:
> I guess we haven't added overlays because me and zimbatm, while often used, we don't think its more readable/cleaner to use overlays. It makes it hard to track in bigger projects where packages are coming from because everything is merged into one big set. Also it has a measurable performance impact to add overlays as it adds another fixed point computation when accessing anything from pkgs (nix does not have lazy data structures so any overlay forces also a copy of your big package attrset). Overlays seems to be commonly used to adding packages to a NixOS configuration, but usually using specialArgs is a better way to make it obvious where packages are coming from.

@Mic92 
Thanks for the detailed explanation! I'm part of the Japanese Nix community and have only been using Nix for about a year, so I really appreciate learning about these considerations — especially the performance implications of overlays and the specialArgs alternative. Great insights from someone on the Nix team. Very helpful!

flake.nix
zimbatm: nope :)

overlays/default.nix
```diff
@@ -0,0 +1,6 @@
+{
+  packages,
+}:
+final: _prev: {
+  llm-agents = packages.${final.stdenv.hostPlatform.system} or { };
```
zimbatm, line 5: People that use overlays typically expect the packages to be evaluated in the context of their own nixpkgs. This re-exports the packages evaluated with this flake's nixpkgs instance, and then attaches them into the user's nixpkgs instance, causing 2 evaluations of nixpkgs.

Issue #1989, opened 2026-01-19
gemini-cli fails to build


rytswd:
### Package

gemini-cli

### Issue type

Build failure

### Platform

x86_64-linux

### Description

Gemini CLI fails to build due to npm hash mismatch

### Error output / logs

```text
gemini-cli-0.24.0-npm-deps> building '/nix/store/fla37m1yzkwva8s0kdwjng58wql88psn-gemini-cli-0.24.0-npm-deps.drv'
gemini-cli-0.24.0-npm-deps> Running phase: unpackPhase
gemini-cli-0.24.0-npm-deps> unpacking source archive /nix/store/xl9abz7bw4lqvhdxdmjicxfck1dxrn7v-source
gemini-cli-0.24.0-npm-deps> source root is source
gemini-cli-0.24.0-npm-deps> Running phase: patchPhase
gemini-cli-0.24.0-npm-deps> Running phase: updateAutotoolsGnuConfigScriptsPhase
gemini-cli-0.24.0-npm-deps> Running phase: configurePhase
gemini-cli-0.24.0-npm-deps> no configure script, doing nothing
gemini-cli-0.24.0-npm-deps> Running phase: buildPhase
gemini-cli-0.24.0-npm-deps> Running phase: fixupPhase
gemini-cli-0.24.0-npm-deps> shrinking RPATHs of ELF executables and libraries in /nix/store/gnrnkanaia68f290zqi8gra1a49a9xd9-gemini-cli-0.24.0-npm-deps
gemini-cli-0.24.0-npm-deps> checking for references to /build/ in /nix/store/gnrnkanaia68f290zqi8gra1a49a9xd9-gemini-cli-0.24.0-npm-deps...
gemini-cli-0.24.0-npm-deps> patching script interpreter paths in /nix/store/gnrnkanaia68f290zqi8gra1a49a9xd9-gemini-cli-0.24.0-npm-deps
error: hash mismatch in fixed-output derivation '/nix/store/fla37m1yzkwva8s0kdwjng58wql88psn-gemini-cli-0.24.0-npm-deps.drv':
         specified: sha256-iw399iCovxDFbfuf0+V4uq9LsKx2VUr1BbriYtJOZiY=
            got:    sha256-hilLONJfva4UP+8pfEnrxypjb8FrYSuiQz0Q1E6S/Ug=
[...]

PR #1990, opened 2026-01-19
gemini-cli: fix npm hash

rytswd:
Fixes #1989

PR #1991, opened 2026-01-19
ci: fix nix-update to verify package build

rytswd:
This is related to #1989, where npm hash mismatch caused the build to fail

I also added 132b527f to update the npm hash when the mismatch is found.

### DIFF
--- a/.github/actions/update.sh
+++ b/.github/actions/update.sh
@@ -60,24 +60,55 @@ if [ "$type" = "package" ]; then
   # Verify the package actually builds with the new hashes
   # This catches npm registry non-determinism issues where hashes become stale
   echo "Verifying package builds..."
-  if ! nix build .#"$name" --no-link 2>&1; then
-    echo "::error::Package $name failed to build after update"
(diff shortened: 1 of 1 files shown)


.github/actions/update.sh
```diff
@@ -55,6 +57,60 @@ if [ "$type" = "package" ]; then
   new_version=$(nix eval .#packages.x86_64-linux."$name".version --raw 2>/dev/null || echo "unknown")
   echo "New version: $new_version"
 
+  # Verify the package actually builds with the new hashes
+  # This catches npm registry non-determinism issues where hashes become stale
+  echo "Verifying package builds..."
+  max_retries=3
+  retry_count=0
+
+  while ! build_output=$(nix build .#"$name" --no-link 2>&1); do
```
Mic92, line 66: Given we have a cache, this won't actually detect issues like: https://github.com/numtide/llm-agents.nix/pull/1990

We also commonly test builds with buildbot when we merge automatically pull requests rather than in github with limited resources on the actions runner.

Mic92:
Given that we already test builds in buildbot, I don't think these changes are needed.

PR #2006, opened 2026-01-20
 prefetch-npm-deps: exclude deprecated field from packument whitelist

Mic92:


multivac61:
I'm having this issue with claude-code-acp too
CleanShot 2026-01-20 at 08 27 38@2x

Mic92:
> I'm having this issue with claude-code-acp too CleanShot 2026-01-20 at 08 27 38@2x

That's definitly not the issue I am fixing here.

Mic92:
I don't know what this issue is, but it almost looks like a broken store path to me. You can use the `nix-store` command to check if the hash of this store path is still intact. Garbage collection might be able to remove it.

PR #2117, opened 2026-01-26
fix: use warnOnInstantiate  for deprecation notice

### LINKED ISSUE #2116: lib.warn in claude-code-npm breaks evaluation with abort-on-warn = true
The deprecation warning in `packages/claude-code-npm/default.nix` uses `lib.warn`:

```nix
pkgs.lib.warn "claude-code-npm is deprecated, use claude-code instead" perSystem.self.claude-code
```

This breaks flake evaluation for users with `abort-on-warn = true` in their nixConfig, even when they never reference `claude-code-npm`.

## Why this happens

Flake-parts internally uses `filterAttrs` to build the packages output, which forces evaluation of all package attributes. Since `lib.warn` executes during evaluation (not when the package is built or used), the warning fires immediately and triggers the abort.

## Reproduction

```nix
{
  nixConfig.abort-on-warn = true;
  
  inputs.llm-agents.url = "github:numtide/llm-agents.nix";
  
  outputs = { llm-agents, ... }: {
    # This fails even though we're accessing spec-kit, not claude-code-npm
    packages.x86_64-linux.default = llm-agents.packages.x86_64-linux.spec-kit;
  };
}
```

## Suggested fix

Use `lib.warnOnInstantiate` which wraps the derivation so the warning only fires when the package is actually **used** (instantiated), not during attrset evaluation:

```nix
pkgs.lib.warnOnInstantiate "claude-code-npm is deprecated, use claude-code instead" (
  perSystem.self.claude-code.overrideAttrs (old: {
    passthru = (old.passthru or { }) // {
      hideFromDocs = true;
    };
  })
)
```

Bad3r:
## Summary
- Use `lib.warnOnInstantiate` instead of `lib.warn` for the claude-code-npm deprecation notice

## Problem
`lib.warn` triggers `abort-on-warn` during attrset evaluation, breaking flake evaluation for users with strict nixConfig settings, even when they don't reference the deprecated package.

Flake-parts uses `filterAttrs` internally to build the packages output, which forces evaluation of all package attributes. Since `lib.warn` executes during evaluation, the warning fires immediately and triggers the abort.

## Solution
`lib.warnOnInstantiate` wraps the derivation so the warning only fires when the package is actually **used** (instantiated), not during attrset evaluation. This is the correct pattern for deprecation warnings on packages.

Fixes #2116

## Test plan
- [x] Verify flake evaluates with `abort-on-warn = true` when accessing other packages (no warning)
- [x] Verify warning still appears when `claude-code-npm` is actually used

### DIFF
--- a/packages/claude-code-npm/default.nix
+++ b/packages/claude-code-npm/default.nix
@@ -1,8 +1,8 @@
 { pkgs, perSystem, ... }:
-(pkgs.lib.warn "claude-code-npm is deprecated, use claude-code instead" perSystem.self.claude-code)
-.overrideAttrs
-  (old: {
+pkgs.lib.warnOnInstantiate "claude-code-npm is deprecated, use claude-code instead" (
+  perSystem.self.claude-code.overrideAttrs (old: {
     passthru = (old.passthru or { }) // {
       hideFromDocs = true;
     };
   })
+)


Bad3r:
UPDATE: the current fix has two problems:                                                                                     
  1. It still prints noise for users who don't use claude-code-npm                                                         
  2. Users who DO use claude-code-npm won't get a proper warning

Bad3r:
UPDATE: Updated the fix from `lib.trace` to `lib.warnOnInstantiate`.

`lib.trace` still prints during attrset evaluation even when the package isn't used. It just avoids the `abort-on-warn` trigger. This means users who never reference `claude-code-npm` would still see the deprecation message on every evaluation.

`lib.warnOnInstantiate` is the correct pattern for package deprecations:
- No output during normal evaluation (accessing other packages)
- Warning only appears when `claude-code-npm` is actually instantiated/used
- Properly triggers `abort-on-warn` for users who are still using the deprecated package (giving them a clear signal to migrate)

PR #2159, opened 2026-01-28
feat(packages): add skills-installer

Bad3r:
## Summary

Add [skills-installer](https://github.com/Kamalnrf/claude-plugins/tree/main/packages/skills-installer) - multi-client skill installer from the claude-plugins monorepo.

## Test plan

- [x] `nix build .#skills-installer` succeeds
- [x] `skills-installer --help` runs correctly
- [x] Package updates via `nix-update --flake skills-installer`
- [x] Smoke test via `installCheckPhase` (tool has no `--version` flag)

### DIFF
--- a/packages/skills-installer/package.nix
+++ b/packages/skills-installer/package.nix
@@ -28,7 +28,7 @@ stdenv.mkDerivation rec {
   doInstallCheck = true;
   installCheckPhase = ''
     runHook preInstallCheck
-    $out/bin/skills-installer --help > /dev/null
+    $out/bin/skills-installer --help | grep -q "Install agent skills"
     runHook postInstallCheck
   '';
 


packages/skills-installer/package.nix
```diff
@@ -0,0 +1,48 @@
...
+stdenv.mkDerivation rec {
+  pname = "skills-installer";
+  inherit (versionData) version;
+
+  src = fetchzip {
+    url = "https://registry.npmjs.org/${pname}/-/${pname}-${version}.tgz";
+    inherit (versionData) hash;
+  };
+
+  nativeBuildInputs = [ nodejs ];
```
Mic92: Please add the `versionsCheckHook` same way as all packages to make sure binaries work correctly.

packages/skills-installer/package.nix
```diff
@@ -0,0 +1,48 @@
...
+
+  passthru.category = "Claude Code Ecosystem";
+
+  meta = with lib; {
+    description = "Install agent skills across multiple AI coding clients";
+    homepage = "https://github.com/Kamalnrf/claude-plugins";
+    changelog = "https://github.com/Kamalnrf/claude-plugins/releases";
+    license = licenses.mit;
+    sourceProvenance = with lib.sourceTypes; [ binaryBytecode ];
+    maintainers = with flake.lib.maintainers; [ ypares ];
```
Mic92: Please use your own maintainer handle here.

packages/skills-installer/package.nix
```diff
@@ -0,0 +1,48 @@
...
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/bin
+    cp $src/dist/cli.js $out/bin/skills-installer
+    chmod +x $out/bin/skills-installer
+
+    substituteInPlace $out/bin/skills-installer \
+      --replace-quiet "#!/usr/bin/env node" "#!${nodejs}/bin/node"
```
Mic92: What does the shebang looks like here? `patchShebangs` usually runs on it's own and should catch this.

packages/skills-installer/package.nix
```diff
@@ -0,0 +1,48 @@
...
+  };
+
+  nativeBuildInputs = [ nodejs ];
+
+  installPhase = ''
+    runHook preInstall
+
+    mkdir -p $out/bin
+    cp $src/dist/cli.js $out/bin/skills-installer
+    chmod +x $out/bin/skills-installer
```
Mic92: can done in one line using the `install` command.

packages/skills-installer/package.nix
```diff
@@ -0,0 +1,48 @@
...
+
+let
+  versionData = lib.importJSON ./hashes.json;
+in
+stdenv.mkDerivation rec {
+  pname = "skills-installer";
+  inherit (versionData) version;
+
+  src = fetchzip {
+    url = "https://registry.npmjs.org/${pname}/-/${pname}-${version}.tgz";
```
Mic92, line 14: I think nix-update actually is able to update these type of packages, so the custom update.py might be not needed. Should be tested.

packages/skills-installer/package.nix
Bad3r: Unfortunately, the tool doesn't have a way to check the version.. Similar to `claude-plugins`. Added a test via `installCheckPhase` that runs `--help` instead

packages/skills-installer/package.nix
Bad3r: Shebang is #!/usr/bin/env node. `patchShebangs` can't resolve env node in the sandbox. Updated to use:
```nix
--replace-fail "#!/usr/bin/env node" "#!${nodejs}/bin/node" 
```

This is the same pattern used in Same pattern as `ccusage-codex` and `claude-plugins`

packages/skills-installer/package.nix
Bad3r: Done. Changed to `install -Dm755 dist/cli.js $out/bin/skills-installer`

packages/skills-installer/package.nix
Bad3r: You are correct. I switched to `nix-update` instead. Confirmed that it works as expected by downgrading the package version and running `nix run nixpkgs#nix-update -- --flake skills-installer`. Removed hashes.json and update.py.

packages/skills-installer/package.nix
Mic92: It actually did work. I had to move node to buildInputs.

PR #2487, opened 2026-02-11
beads-rust: init at 0.1.13

afterthought:
## Summary

- Adds `beads-rust` (`br`), a fast Rust port of Steve Yegge's beads — a local-first, non-invasive issue tracker for git repositories
- Source: [Dicklesworthstone/beads_rust](https://github.com/Dicklesworthstone/beads_rust)
- Packaged as binary distribution from GitHub releases
- Platforms: x86_64-linux, x86_64-darwin, aarch64-darwin
- Category: Workflow & Project Management (alongside existing `beads` package)

## Differences from `beads`

The existing `beads` package builds the original Go implementation (`steveyegge/beads`, binary: `bd`). This adds the Rust port (`Dicklesworthstone/beads_rust`, binary: `br`) which provides the same functionality with improved performance.

## Test plan

- [ ] `nix build .#beads-rust` on aarch64-darwin
- [ ] `versionCheckHook` passes (`br --version` outputs `br 0.1.13`)
- [ ] Binary runs correctly: `nix run .#beads-rust -- --help`

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2509, opened 2026-02-12
get-shit-done: init at 1.18.0

ingolf-wagner:
## Summary

add get-shit-done. A workflow, which is kinda a mixture of [ralph](https://ghuntley.com/ralph/) and openspec

## Test plan

> [!NOTE]

### DIFF
--- a/packages/get-shit-done/package.nix
+++ b/packages/get-shit-done/package.nix
@@ -1,0 +1,49 @@
+{
+  lib,
+  buildNpmPackage,
+  fetchFromGitHub,
+}:
+
+buildNpmPackage rec {
+  pname = "get-shit-done";
+  version = "1.18.0";
+
+  src = fetchFromGitHub {
+    owner = "gsd-build";
+    repo = "get-shit-done";
+    rev = "v${version}";
+    hash = "sha256-PbvmJkFv1NHd7pc+N4lVh/8ZiQHuPpUpCZLQIX3VZxs=";
+  };
+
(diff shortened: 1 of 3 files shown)


packages/get-shit-done/package.nix
```diff
@@ -0,0 +1,49 @@
...
+  '';
+
+  passthru.category = "Workflow & Project Management";
+
+  meta = with lib; {
+    description = "Meta-prompting, context engineering and spec-driven development system for Claude Code";
+    homepage = "https://github.com/gsd-build/get-shit-done";
+    license = licenses.mit;
+    sourceProvenance = with lib.sourceTypes; [ fromSource ];
+    maintainers = with maintainers; [ ];
```
Mic92, line 45: Can you add yourself as maintainer, same way as in this PR? https://github.com/numtide/llm-agents.nix/pull/2487

ingolf-wagner: Oh, missed that. To be honest, I don't think this needs to be packaged anymore.

It looks like this project has no dependencies and is plain js.

https://github.com/gsd-build/get-shit-done/blob/main/package.json

when running `npx get-shit-done-cc@latest` in your project it will install a bunch of scripts in `./.claude` or `./.opencode` and in these scripts it runs `npx ./osd-scriopt.js` anyway. So I guess packaging this project is not necessary.

Mic92: brave new world of packaging.... where the package is just build for the side effect.

PR #2541, opened 2026-02-13
feat(auto-claude): adding auto-claude tool

xorilog:
## Summary

This PR adds https://github.com/AndyMik90/Auto-Claude tool
> Autonomous multi-agent coding framework that plans, builds, and validates software for you.

I have an issue as there is no released binary for `aarch64-linux` architecture, i do not know if this is blocking or if i should find an other way to do it... please tell me as i am pretty new in nix !

Also the binary does not support `--version`, i did not used versionCheckHook or versionCheckHomeHook...
## Test plan

- [x] `nix fmt` succeeds.
- [x] `nix build --accept-flake-config .#auto-claude` succeeds.
- [x] Package updates via custom `update.py` works

> [!NOTE]

### DIFF
--- a/packages/auto-claude/package.nix
+++ b/packages/auto-claude/package.nix
@@ -1,0 +1,123 @@
+{
+  lib,
+  flake,
+  stdenv,
+  fetchurl,
+  appimageTools,
+  unzip,
+  makeWrapper,
+}:
+
+let
+  versionData = builtins.fromJSON (builtins.readFile ./hashes.json);
+  inherit (versionData) version hashes;
+
+  pname = "auto-claude";
+
+  platform = stdenv.hostPlatform.system;
+
+  sources = {
+    x86_64-linux = {
+      url = "https://github.com/AndyMik90/Auto-Claude/releases/download/v${version}/Auto-Claude-${version}-linux-x86_64.AppImage";
+      hash = hashes.x86_64-linux;
+    };
+    x86_64-darwin = {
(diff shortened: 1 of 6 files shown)


packages/auto-claude/package.nix
```diff
@@ -0,0 +1,123 @@
...
+  flake,
+  stdenv,
+  fetchurl,
+  appimageTools,
+  unzip,
+  makeWrapper,
+}:
+
+let
+  versionData = builtins.fromJSON (builtins.readFile ./hashes.json);
```
Mic92, line 12: Project seems to use pnpm, could we try to build from source first? we have some other examples in the repo doing the same.

Mic92: @copilot you can try. This would allow us to remove the update.py. Please test the result.

Mic92: Could you test this? https://github.com/numtide/llm-agents.nix/pull/2543

xorilog: It launches as well ! but it opens the developper tools tabs at launch 🤔

Mic92: @xorilog do you know why it depends on electron? This one seems a bit odd. The developer tools can be probably very likely be disabled given proper build flags.

Mic92: Fixed in https://github.com/numtide/llm-agents.nix/pull/2543

Mic92: merged the pr.

Mic92:
Now doing: https://github.com/numtide/llm-agents.nix/pull/2543 instead

PR #2624, opened 2026-02-18
ci: run update every 2 hours

andresilva:
## Summary

Run the update CI workflow every two hours. Not sure if there was a reason for just doing it twice daily? The job is relatively fast, and it's an open-source repo so limits shouldn't be a problem.

### DIFF
--- a/.github/workflows/update.yml
+++ b/.github/workflows/update.yml
@@ -1,8 +1,8 @@
 name: Update Dependencies
 on:
   schedule:
-    # Run twice daily at 2 AM and 2 PM UTC
-    - cron: '0 2,14 * * *'
+    # Run every 2 hours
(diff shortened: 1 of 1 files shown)


Mic92:
Sorry, but this interferes more with me doing updates. I often assign copilot to open pull requests that fail to update and having the bot running even more often, makes its more likely to run into git conflicts. Updating 2 times a day should be enough for now. LLM agents update often, but not that often.

andresilva:
It's not so much about release about cadence but more about delay, if this was run every 2 hours most of the runs would have nothing to update, but the delay between a new release and the update would be at most 2 hours, whereas now it's 12. Either way it's understandable if this disrupts your workflow.

PR #2696, opened 2026-02-21
feat: add gno package (local knowledge engine)

afterthought:
## Summary

- Packages [GNO](https://github.com/gmickel/gno) (`@gmickel/gno`) v0.16.0 as a Nix derivation
- GNO is a local-first knowledge engine with hybrid search (BM25 + vector + reranking), RAG Q&A with citations, and an MCP server for AI agent integration
- Follows the `qmd` `buildNpmPackage` pattern: Bun runtime, node-llama-cpp native bindings, sqlite, detectGlibc patch
- Includes GPU support (CUDA/Vulkan) on Linux
- Includes `update.py` for version bumping

## Verification

Tested on `aarch64-darwin`:
- `nix build .#gno` succeeds
- `gno --help` shows usage
- `gno --version` reports `0.16.0`
- `gno mcp --help` shows MCP server commands

## Files

- `packages/gno/default.nix` — callPackage wrapper (same as qmd)
- `packages/gno/package.nix` — buildNpmPackage derivation
- `packages/gno/package-lock.json` — generated from GNO v0.16.0 source
- `packages/gno/node-llama-cpp-detectGlibc.patch` — copied from qmd
- `packages/gno/update.py` — version update script

### DIFF
--- a/packages/gno/update.py
+++ b/packages/gno/update.py
@@ -35,7 +35,8 @@ def main() -> None:
     data = load_hashes(HASHES_FILE)
     current_version = data["version"]
 
-    tag, latest_version = fetch_github_latest_release(OWNER, REPO)
+    latest_version = fetch_github_latest_release(OWNER, REPO)
+    tag = f"v{latest_version}"
     print(f"Current version: {current_version}")
     print(f"Latest version:  {latest_version} ({tag})")
 
@@ -50,7 +51,8 @@ def main() -> None:
     extract_or_generate_lockfile(
         tarball_url=tarball_url,
         output_path=SCRIPT_DIR / "package-lock.json",
-        extra_npm_flags=["--legacy-peer-deps"],
+        # Use legacy-peer-deps to resolve peer dependency conflicts
+        env={"NPM_CONFIG_LEGACY_PEER_DEPS": "true"},
     )
 
     # Calculate source hash


packages/gno/package.nix
```diff
@@ -0,0 +1,167 @@
...
+  # Vulkan supported on all Linux
+  effectiveVulkanSupport = vulkanSupport && stdenv.isLinux;
+
+  version = "0.16.0";
+
+  src = fetchFromGitHub {
+    owner = "gmickel";
+    repo = "gno";
+    tag = "v${version}";
+    hash = "sha256-PPTjQNFABtAosfF6lTwxN14ce2SsbKNG0tXWS6vBLrI=";
```
Mic92: The update.py should generate a hashes.json file that gets imported here. This is much more robust than patching this nix file.

packages/gno/package.nix
```diff
@@ -0,0 +1,167 @@
...
+  };
+
+  # Add package-lock.json (upstream uses bun.lockb)
+  postPatch = ''
+    cp ${./package-lock.json} package-lock.json
```
Mic92, line 40: Could you open also an issue upstream to maintain a lockfile since this an app and not a library?

afterthought:
Sorry I was out for a few days. I will check on this over the weekend.

packages/gno/package.nix
afterthought: Done — moved version, srcHash, and npmDepsHash into `hashes.json`, and `package.nix` now reads them via `builtins.fromJSON (builtins.readFile ./hashes.json)`. The `update.py` uses the shared `scripts/updater` library (`load_hashes`/`save_hashes`) instead of regex-patching.

packages/gno/package.nix
afterthought: GNO already maintains a `bun.lock` (text-format, Bun v1.2+). We explored using [bun2nix](https://github.com/nix-community/bun2nix) to consume it directly, which would eliminate the need for a generated `package-lock.json`. However, we hit a macOS sandbox issue: bun2nix's `extract-bun-package` script uses `bsdtar` which fails on packages with non-ASCII filenames (e.g. `@fastify/send` contains `snow ☃/index.html`) when `LANG` is unset in the Nix sandbox.

Keeping the generated `package-lock.json` approach for now. Will revisit bun2nix once the locale issue is resolved upstream.

PR #3185, opened 2026-03-13
hermes-agent: init at 2026.3.12

aliez-ren:
## Summary
- Add [Hermes Agent](https://hermes-agent.nousresearch.com/) — a self-improving AI agent by Nous Research that creates skills from experience and runs anywhere
- Built from source using `buildPythonApplication` with setuptools; includes inline `fal-client` package (not yet in nixpkgs)
- Patches upstream `pyproject.toml` to fix missing packages for non-editable installs (`agent`, `hermes_state`, `hermes_time`)
- Adds `nix-update-args` with `--version-regex` for date-based tag format (`v2026.3.12`)

## Test plan
- [x] `nix build .#hermes-agent` succeeds
- [x] `hermes --version` outputs `Hermes Agent v0.2.0 (2026.3.12)`
- [x] `nix fmt` passes (0 changed)
- [x] Version check hook passes (matches `2026.3.12` in output)
- [x] `nix flake check` (CI)

```
$ ./result/bin/hermes --version
Hermes Agent v0.2.0 (2026.3.12)
Project: /nix/store/afg6brds7c545hzbfnqd67j1s3l1kila-hermes-agent-2026.3.12/lib/python3.13/site-packages
Python: 3.13.12
OpenAI SDK: 2.21.0
```

### DIFF
--- a/packages/hermes-agent/package.nix
+++ b/packages/hermes-agent/package.nix
@@ -3,6 +3,7 @@
   flake,
   python3,
   fetchFromGitHub,
+  fetchpatch,
   fetchPypi,
   versionCheckHook,
   versionCheckHomeHook,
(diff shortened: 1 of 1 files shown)


packages/hermes-agent/package.nix
```diff
@@ -0,0 +1,132 @@
...
+  versionCheckProgramArg = [ "--version" ];
+
+  passthru.category = "AI Assistants";
+
+  meta = with lib; {
+    description = "Self-improving AI agent by Nous Research — creates skills from experience and runs anywhere";
+    homepage = "https://hermes-agent.nousresearch.com/";
+    changelog = "https://github.com/NousResearch/hermes-agent/releases/tag/v${version}";
+    license = licenses.mit;
+    sourceProvenance = with sourceTypes; [ fromSource ];
```
Mic92, line 122: can you add yourself as a maintainer here?

packages/hermes-agent/package.nix
```diff
@@ -0,0 +1,132 @@
...
+
+  src = fetchFromGitHub {
+    owner = "NousResearch";
+    repo = "hermes-agent";
+    rev = "v${version}";
+    hash = "sha256-zOQm9cVjtwpDteIjjQsWn3SfCfw9tobtoAPU65rLYWA=";
+  };
+
+  postPatch = ''
+    # The upstream pyproject.toml is incomplete for non-editable installs:
```
Mic92: do you think upstream would also accept these patches?

aliez-ren: I'm not a python developer, here is the answer from AI:

> This is a common issue with projects that rely on editable installs (pip install -e .) during development — in editable mode, Python finds modules directly from the source tree, so incomplete pyproject.toml configuration goes unnoticed. Only when doing a proper non-editable install (as Nix does) do these omissions surface.

Mic92: Just give it a try, I will not block the PR on this, but it will help to make the build instructions shorter and avoid breakages in the future.

Mic92: Fixes are actually already upstream. I just included the patch instead.

aliez-ren: wow, great

PR #3355, opened 2026-03-20
amp-bin: init at 0.0.1773965070-g233ce5

winterqt:
## Summary

This is Amp's binary release, which has the advantage of being the preferred installation method as well as the only way to use the plugin API [0].

For simplicity, we piggyback off of the main Amp package's update script.

[0]: https://ampcode.com/manual/plugin-api

## Test plan

- [x] `nix build .#<package>` succeeds (on `aarch64-{darwin,linux}`)

### DIFF
--- a/packages/amp/package.nix
+++ b/packages/amp/package.nix
@@ -5,6 +5,9 @@
   makeWrapper,
   wrapBuddy,
   ripgrep,
+  cctools,
+  darwin,
+  rcodesign,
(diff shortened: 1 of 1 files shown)


Mic92:
I think for now, I rather make the default amp output match upstream in terms of features. Haven't checked if bun2nix would allow the source build. However I was able to drop the __noChroot flag.

PR #3647, opened 2026-04-01
Init nono

pogoba:
## Summary

Add nono package, the coolest LLM sandbox on the market (because i want to contribute to it).

## Test plan


pogoba:
re: build CI failures: hashes seem fine on my end. Maybe github is handing out different data to other users again. I will re-trigger tomorrow or so.

PR #3936, opened 2026-04-12
oh-my-opencode: 3.16.0 -> 3.17.0, fix stale bun.lock

noamsto:
## Summary

v3.17.0 ships a bun.lock still referencing 3.16.0 platform binaries. Add fix-stale-bun-lock.patch to correct it at build time.

Also update scripts/updater/bun.py to auto-generate this patch in future updates when pkg_dir is provided, instead of raising an error.

## Test plan

### DIFF
--- a/packages/oh-my-opencode/fix-stale-bun-lock.patch
+++ b/packages/oh-my-opencode/fix-stale-bun-lock.patch
@@ -1,0 +1,62 @@
+diff --git a/bun.lock b/bun.lock
+index 32e143f..f7e37f7 100644
+--- a/bun.lock
++++ b/bun.lock
+@@ -30,17 +30,17 @@
+         "typescript": "^5.7.3",
+       },
+       "optionalDependencies": {
+-        "oh-my-opencode-darwin-arm64": "3.16.0",
+-        "oh-my-opencode-darwin-x64": "3.16.0",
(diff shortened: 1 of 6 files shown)


packages/oh-my-opencode/fix-stale-bun-lock.patch
```diff
@@ -0,0 +1,62 @@
...
+-        "oh-my-opencode-darwin-x64-baseline": "3.16.0",
+-        "oh-my-opencode-linux-arm64": "3.16.0",
+-        "oh-my-opencode-linux-arm64-musl": "3.16.0",
+-        "oh-my-opencode-linux-x64": "3.16.0",
+-        "oh-my-opencode-linux-x64-baseline": "3.16.0",
+-        "oh-my-opencode-linux-x64-musl": "3.16.0",
+-        "oh-my-opencode-linux-x64-musl-baseline": "3.16.0",
+-        "oh-my-opencode-windows-x64": "3.16.0",
+-        "oh-my-opencode-windows-x64-baseline": "3.16.0",
++        "oh-my-opencode-darwin-arm64": "3.17.0",
```
Mic92, line 20: has upstream also been informed of that?

noamsto: I haven't opened an issue there yet, will do

PR #4000, opened 2026-04-14
omp: resolve bun workspace catalog refs for 14.1.1

DavHau:
Fixes the build for omp 14.1.1 (#3997).

Upstream introduced bun workspace `catalog:` references in root `package.json`. bun tries to resolve these against the npm registry at install/compile time, which fails in the Nix sandbox:

```
error: @anthropic-ai/sdk@catalog: failed to resolve
error: @biomejs/biome@catalog: failed to resolve
...
```

This patch adds a `postPatch` step that inlines every `catalog:` reference from the root `package.json` catalog into all workspace `package.json` files and `bun.lock` before bun runs, so no registry lookup is needed.

Verified with `nix build .#omp`; `omp --version` reports `omp/14.1.1`.


Mic92:
I am fixing this upstream: https://github.com/nix-community/bun2nix/pull/86

Mic92:
Fixed in https://github.com/numtide/llm-agents.nix/pull/3997

PR #4093, opened 2026-04-18
fence: init at 0.1.48

uesyn:
## Summary

- Add a package [Use-Tusk](https://github.com/Use-Tusk)[fence](https://github.com/Use-Tusk/fence) — Lightweight, container-free sandbox for running commands with network and filesystem restrictions

## Test plan

- [x] `nix build .#fence -- --version` succeeds


packages/fence/package.nix
```diff
@@ -0,0 +1,85 @@
...
+
+  passthru.category = "Utilities";
+
+  meta = with lib; {
+    description = "Lightweight, container-free sandbox for running commands with network and filesystem restrictions";
+    homepage = "https://fencesandbox.com/";
+    changelog = "https://github.com/Use-Tusk/fence/releases";
+    license = licenses.asl20;
+    sourceProvenance = with sourceTypes; [ fromSource ];
+    maintainers = with maintainers; [ ];
```
uesyn: Should I list myself as the maintainer?

packages/fence/package.nix
```diff
@@ -0,0 +1,85 @@
...
+
+  passthru.category = "Utilities";
+
+  meta = with lib; {
+    description = "Lightweight, container-free sandbox for running commands with network and filesystem restrictions";
+    homepage = "https://fencesandbox.com/";
+    changelog = "https://github.com/Use-Tusk/fence/releases";
+    license = licenses.asl20;
+    sourceProvenance = with sourceTypes; [ fromSource ];
+    maintainers = with maintainers; [ uesyn ];
```
Mic92: You need to import the flake attribute and than use flake.maintainers here. Check for other packages in the repo.

Mic92:
Pushed a fixup: `package.nix` is invoked via `pkgs.callPackage`, so `lib.maintainers` resolves to nixpkgs' set, not the one extended in `lib/default.nix`. Passed `flake` through and switched to `flake.lib.maintainers` (same pattern as `packages/bernstein`).

PR #4322, opened 2026-04-25
claude-code-acp: rename to claude-agent-acp

kamalmarhubi:
## Summary

Upstream renamed and moved orgs:
- agentclientprotocol/claude-agent-acp#319
- agentclientprotocol/claude-agent-acp#471

The old name is kept as a deprecated alias with a warning.

## Test plan


kamalmarhubi:
re maintainer check failure: the original maintainer list was empty, I just did a straight rename. Guessing the check is newer than that package?

PR #4433, opened 2026-04-29
ci: fail PRs when README is out of date

sei40kr:
## Summary
- Add `.github/workflows/check-readme.yml` that runs on `pull_request` (and `workflow_dispatch`).
- Regenerates the README via `./scripts/generate-package-docs.py` and fails the job if `git diff README.md` is non-empty, prompting the contributor to commit the regenerated docs.

## Test plan
- [x] Open a PR that touches a package's metadata without regenerating the README and confirm the workflow fails with the README-out-of-date error.
- [x] Open a PR with the regenerated README committed and confirm the workflow passes.

### DIFF
--- a/.github/workflows/check-readme.yml
+++ b/.github/workflows/check-readme.yml
@@ -1,0 +1,22 @@
+name: Check README is up to date
+on:
+  pull_request:
+    paths:
+      - 'packages/**'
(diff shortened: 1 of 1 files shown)


Mic92:
I will add this, but not so sure if I want to make this a required check yet. Shouldn't be super important for package updates.

PR #4470, opened 2026-04-30
add the td tool as an alternative option to beads

mbainter:
## Summary

Adds the [td](https://github.com/marcus/td) tool as an alternative to beads

## Test plan

Ran the checks and tested the nix build.

- [X] `nix build .#<package>` succeeds
- [X] Package updates via `nix-update` or has a custom `update.py` if nix-update doesn't work

I am not particularly familiar with nix-update, but this is a straightforward go package. Based on other similar go implementations in this repository that apparently work with nix-update, and a quick perusal of its docs I expect this should work the same as those.

### DIFF
--- a/README.md
+++ b/README.md
@@ -717,6 +717,16 @@ Nix packages for AI coding agents and development tools. Automatically updated d
 
 </details>
 <details>
+<summary><strong>td</strong> - A minimalist CLI for tracking tasks across AI coding sessions.</summary>
+
+- **Source**: source
+- **License**: MIT
+- **Homepage**: https://github.com/marcus/td
+- **Usage**: `nix run github:numtide/llm-agents.nix#td -- --help`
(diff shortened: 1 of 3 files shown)


Mic92:
Is nix-update the only reason you marked this as draft?

mbainter:
Yeah, I was going to make sure the checks all passed and come back to see if I could test the nix-update flow.  Unfortunately I've been sick and haven't been able to come back to it.  I marked it ready, in case you're asking because you're comfortable with it as-is.

Mic92:
Tested `nix-update` locally (downgraded to 0.43.0, then ran `nix run nixpkgs#nix-update -- --flake td`): it correctly bumped version, src hash, and vendor hash to 0.44.0. Package is ready for automated updates as-is.

— posted by an automated assistant on @Mic92's behalf

PR #4501, opened 2026-04-30
agentsview: restore update.py — wesm/agentsview#429 fixed regression

ak2k:
The skip stub from #4477 has done its job. [wesm/agentsview#429](https://github.com/wesm/agentsview/pull/429) (merged) fixes the timing-endpoint nil-slice bug that caused the 0.26.0 frontend to throw on every session view.

Once a release tag containing the fix is cut, the daily updater should pick it up; restoring \`update.py\` unblocks that. \`hashes.json\` stays at 0.25.0 — the next successful run of the updater bumps it.

### DIFF
--- a/packages/agentsview/update.py
+++ b/packages/agentsview/update.py
@@ -1,8 +1,76 @@
-#!/usr/bin/env python3
-"""Prevent automated agentsview bumps while 0.26.0 suffers a frontend regression.
+#!/usr/bin/env nix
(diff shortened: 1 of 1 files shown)


ak2k:
Closing — was premature. \`fetch_github_latest_release\` would return v0.26.0 (still latest tag) and \`should_update("0.25.0", "0.26.0")\` is true, so this would bump us straight back into the regression. Will reopen with a guard, or after wesm tags a release > 0.26.0.

PR #4563, opened 2026-05-03
vibe-kanban: add update.py

zitongcharliedeng:
vibe-kanban shipped without an update.py so the daily auto-bump workflow has skipped it since 2026-02-06. Adds one mirroring the existing pattern (codex/agentsview/cc-sdd/etc.).

Behavior: fetch latest BloopAI/vibe-kanban release tag, derive semver, compute hashes for src tarball, release ZIP (where the prebuilt frontend with the react-virtuoso license key lives), cargoHash, npmDepsHash. Writes hashes.json.

Tested locally on a clean 0.1.2 hashes.json: `nix shell --inputs-from .# nixpkgs#python3 nixpkgs#nix-prefetch-scripts -c python3 packages/vibe-kanban/update.py` runs deterministically to v0.1.44.

Bump itself is left to the auto-update workflow rather than included here, since v0.1.44 reshapes the source tree (pnpm workspace under `packages/`) and `package.nix`'s `cd frontend` step needs a separate adjustment. Splitting that into its own PR keeps this one focused on the missing convention.

### DIFF
--- a/packages/vibe-kanban/update.py
+++ b/packages/vibe-kanban/update.py
@@ -1,0 +1,96 @@
+#!/usr/bin/env nix
+#! nix shell --inputs-from .# nixpkgs#python3 --command python3
+
+"""Update script for vibe-kanban package."""
+
+import json
+import sys
+import urllib.request
+from pathlib import Path
+
+sys.path.insert(0, str(Path(__file__).parent.parent.parent / "scripts"))
+
+from updater import (
+    calculate_dependency_hash,
+    calculate_url_hash,
+    load_hashes,
+    save_hashes,
+    should_update,
+)
+from updater.hash import DUMMY_SHA256_HASH
+from updater.nix import NixCommandError, nix_prefetch_url
+
+HASHES_FILE = Path(__file__).parent / "hashes.json"
+
+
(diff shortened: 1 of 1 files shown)


zitongcharliedeng:
Aside, separate from this PR.

scripts/updater plus 43 per-package update.py scripts is roughly 4,300 LOC of Python whose net effect is fetch latest tag, compute four hashes, write JSON. Clean inside the repo, but grows O(packages).

Two prior-art options:
- nvfetcher: single TOML per source, nix-native, generates _sources/generated.{nix,json}.
- Dependabot `package-ecosystem: nix`: real but appears scoped to flake.lock; would not cover hashes.json files inside derivations here.

Was migration to nvfetcher (or equivalent) considered? Given numtide's track record on devshell, treefmt-nix, srvos, blueprint, etc. I'd love to read the design rationale before submitting more package additions. Per-package quirks like vibe-kanban's release-zip license-key extraction may push back against a generic tool, in which case noting that somewhere visible would help future contributors.

Mic92:
Superseded by #4611, which merged the structural 0.1.2 → 0.1.44 bump together with a tightened version of this update.py. Thanks for the original groundwork — your script is what got tested and shipped.

PR #4568, opened 2026-05-03
vibe-kanban: 0.1.2 -> 0.1.44 + package.nix layout fix

zitongcharliedeng:
Stacked on top of #4563 (which adds update.py).

Once #4563 merges, the daily auto-update workflow will run update.py and produce the 0.1.2 -> 0.1.44 hashes.json bump on its own. But that auto-PR will fail nix-build because v0.1.44 reshapes the source tree. This PR carries the package.nix changes needed for v0.1.44 to actually build.

Layout changes vs v0.1.2:

- Frontend moved into a pnpm workspace package at `packages/local-web` (was top-level `frontend/`). Updated buildPhase `cd packages/local-web` and Rust preBuild `mkdir -p packages/local-web/dist`. crates/server/src/routes/frontend.rs uses rust-embed with folder `../../packages/local-web/dist`.
- The mcp binary is now named directly: crates/mcp produces a `[[bin]] name = "vibe-kanban-mcp"` (was `mcp_task_server` then renamed in postInstall). Added `--package mcp` to cargoBuildFlags and dropped the rename in postInstall.
- A transitive dep activates openssl-sys with the `vendored` feature, which would compile OpenSSL from source via openssl-src (needs perl). Set `OPENSSL_NO_VENDOR=1` so the build links dynamically against the openssl already in buildInputs.

Local nix-build is green; all three binaries (vibe-kanban 153MB, vibe-kanban-mcp 19MB, vibe-kanban-review 11MB) produce and run.

Marking draft so #4563 lands first; will mark ready immediately after that merges.

### DIFF
--- a/packages/vibe-kanban/hashes.json
+++ b/packages/vibe-kanban/hashes.json
@@ -1,8 +1,8 @@
 {
-  "version": "0.1.2",
-  "tag": "v0.1.2-20260203101746",
-  "hash": "sha256-rC9r9UPkI75vf4Fk1snBz5KDf/U8lw6XrO2nF8gplpo=",
-  "cargoHash": "sha256-IIpLTlHfDhMTwaUwNEGECQXnsWmPpibt6FkO8smrnBA=",
-  "npmDepsHash": "sha256-wY/ATOO3OcUc72ScmHradTnwBG7PZOqIFrTdHSAAY+8=",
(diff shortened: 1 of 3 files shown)


Mic92:
Superseded by #4611, which merged the structural 0.1.2 → 0.1.44 bump together with a tightened version of this update.py. Thanks for the original groundwork — your script is what got tested and shipped.

PR #4590, opened 2026-05-04
Update README.md binary cache to match unstable

gamedolphin:
## Summary

extra-substitutors is not a thing in the docs at least - https://search.nixos.org/options?channel=unstable&query=substituters&show=option:nix.settings.trusted-substituters

## Test plan

Nope. Just a readme.md update.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1118,8 +1118,8 @@ Alternatively, you can configure this system-wide in your NixOS configuration:
 
 ```nix
 nix.settings = {
-  extra-substituters = [ "https://cache.numtide.com" ];
-  extra-trusted-public-keys = [
+  substituters = [ "https://cache.numtide.com" ];
+  trusted-public-keys = [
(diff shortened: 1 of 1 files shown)


Mic92:
`extra-*` is a generic Nix mechanism, not a per-option declaration: any list-valued setting automatically gets an `extra-` form that appends instead of replaces. See [`src/libutil/configuration.cc#L22-L39`](https://github.com/NixOS/nix/blob/fc08c86a077e4bd0ca77a161b90a4b2747437baa/src/libutil/configuration.cc#L22-L39):

```cpp
if (hasPrefix(name, "extra-")) {
    i = _settings.find(std::string(name, 6));
    if (i == _settings.end() || !i->second.setting->isAppendable())
        return false;
    append = true;
}
```

So `extra-substituters` / `extra-trusted-public-keys` are valid and are what you want here — using bare `substituters` would drop `cache.nixos.org` from the user's config.

gamedolphin:
TIL! Thanks for the clarification! Cheers! Updating my own config right now!

PR #4913, opened 2026-05-15
grok: wrap with bubblewrap to provide /bin shells on NixOS

### LINKED ISSUE #4912: grok: Cannot execute any commands
### Package

grok

### Issue type

Runtime error

### Platform

x86_64-linux

### Nix sandbox enabled?

Unknown

### Description

grok fails to execute all commands. This is because grok's run_command tool directly calls execve on /bin/<basename($SHELL)> (or /bin/bash as a fallback). Since NixOS only provides /bin/sh and lacks both /bin/bash and /bin/zsh, the execution fails with ENOENT.

Image

  Verification using strace:
  - $SHELL=...zsh → execve("/bin/zsh", ["-c", "pwd"]) → ENOENT
  - $SHELL=...bash or fallback → execve("/bin/bash", ...) → ENOENT

In NixOS, /bin/zsh is not provided by default and you must respect the $SHELL environment variable.
By packaging grok to properly honor $SHELL, it will function correctly in NixOS environments.

### Error output / logs

```text

```

conao3:
## Summary

Fixes #4912.

On NixOS (and other non-FHS Linux distros) the packaged `grok` binary fails every `run_command` tool call with:

```
Terminal error: IO Error: No such file or directory (os error 2)
```

The Grok CLI's `run_command` uses `portable_pty::cmdbuilder` which `execve`s `/bin/<basename($SHELL)>` directly (e.g. `/bin/bash`, `/bin/zsh`). On NixOS only `/bin/sh` exists, so the bash tool, git status, pwd, etc. all return `ENOENT` before any user command runs.

Wrap the Linux `grok` and `agent` entrypoints with `bubblewrap`: place `makeWrapper` output under `libexec/grok/{grok,agent}-launcher` and write a shell wrapper at `bin/{grok,agent}` that runs the launcher inside a bwrap namespace with `/bin` replaced by a tmpfs containing symlinks to `bashInteractive/bin/{bash,sh}` and `zsh/bin/zsh` from the Nix store. macOS keeps the previous `makeWrapper`-only behavior (symlinked into `bin/`).

The `--no-auto-update` flag and `--argv0 {grok,agent}` semantics are preserved by keeping the existing `makeWrapper` calls — they now produce the launchers that the bwrap wrapper exec's into.

After appling this patch, my grok can execute any commands.

image

---

Problem investigation by ClaudeCode and its response. (just FYI; [requesting by maintainer](https://x.com/ryoppippi/status/2055283405303009462))

HIXHOD-b0AAG4v3

## Test plan

- `nix build .#grok` succeeds (versionCheckHook runs `grok --version` through the bwrap wrapper successfully)
- On NixOS, `grok --always-approve` in a tmux pane runs `pwd` and `git status` via the bash tool and returns the expected output (previously every shell tool invocation failed with the `os error 2` shown above)
- Headless mode: `grok -p "Run pwd via bash" --always-approve --output-format json` returns `/home/.../current-dir` instead of the terminal error

> [!NOTE]

---
Assisted-by: Claude Code:claude-opus-4-7[1m]

### DIFF
--- a/packages/grok/package.nix
+++ b/packages/grok/package.nix
@@ -77,6 +77,7 @@ stdenv.mkDerivation {
     done
   ''
   + lib.optionalString (!stdenv.hostPlatform.isLinux) ''
+    install -d $out/bin
     ln -s $out/libexec/grok/grok-launcher $out/bin/grok
     ln -s $out/libexec/grok/agent-launcher $out/bin/agent
   ''
--- a/packages/grok/package.nix
+++ b/packages/grok/package.nix
@@ -6,6 +6,9 @@
   wrapBuddy,
   versionCheckHook,
   versionCheckHomeHook,
+  bashInteractive,
+  bubblewrap,
+  zsh,
 }:
 
 let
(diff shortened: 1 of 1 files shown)


ryoppippi (approved):
lgtm

conao3:
Thanks!

PR #5044, opened 2026-05-19
antigravity: init at 1.0.0

ryoppippi:
## Summary

Add a package for Google Antigravity CLI at 1.0.0. The package uses the official per-platform release manifests and installs the upstream binary as agy, with antigravity as a convenience symlink.

## What Changed

- Added packages/antigravity with fixed manifest URLs and SHA512 hashes for Linux and Darwin on x86_64/aarch64.
- Added a custom updater that reads the upstream manifest endpoint.
- Regenerated the README package list.

## Testing

- nix fmt -- packages/antigravity
- nix build --accept-flake-config path:.#antigravity
- nix run --accept-flake-config path:.#antigravity -- --version
- ./result/bin/agy --version
- ./result/bin/antigravity --version
- uv run python packages/antigravity/update.py
- nix build --accept-flake-config --dry-run --system x86_64-linux path:.#antigravity

### DIFF
--- a/packages/antigravity/package.nix
+++ b/packages/antigravity/package.nix
@@ -1,0 +1,61 @@
+{
+  lib,
+  stdenv,
+  fetchurl,
+  wrapBuddy,
+  versionCheckHook,
+  versionCheckHomeHook,
+}:
+
+let
+  versionData = builtins.fromJSON (builtins.readFile ./hashes.json);
+  inherit (versionData) version urls hashes;
+
+  platform = stdenv.hostPlatform.system;
+in
+stdenv.mkDerivation {
+  pname = "antigravity";
+  inherit version;
+
+  src = fetchurl {
+    url = urls.${platform} or (throw "Unsupported system: ${platform}");
+    hash = hashes.${platform} or (throw "Unsupported system: ${platform}");
+  };
+
+  nativeBuildInputs = lib.optionals stdenv.hostPlatform.isLinux [ wrapBuddy ];
+
+  sourceRoot = ".";
+
+  installPhase = ''
+    runHook preInstall
+
+    install -Dm755 antigravity $out/bin/agy
+    ln -s agy $out/bin/antigravity
+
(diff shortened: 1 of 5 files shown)


packages/antigravity/package.nix
```diff
@@ -0,0 +1,75 @@
...
+stdenv.mkDerivation {
+  pname = "antigravity";
+  inherit version;
+
+  src = fetchurl {
+    url = urls.${platform} or (throw "Unsupported system: ${platform}");
+    hash = hashes.${platform} or (throw "Unsupported system: ${platform}");
+  };
+
+  nativeBuildInputs = lib.optionals stdenv.hostPlatform.isLinux [ autoPatchelfHook ];
```
Mic92, line 25: Would `wrapBuddy` work here? It's less invasive when it comes to binary patching.

ryoppippi: I tried `wrapBuddy` first, It built far enough to patch the Linux binaries, but the patched `agy` segfaulted during the install check

ryoppippi:
@Mic92 now it works on linux as well(i tested on mac but now it works on nixos as well)

packages/antigravity/package.nix
Mic92: Interesting. First segfault that is caused by wrapBuddy. Maybe some library causing the issue here.

ryoppippi: maybe we can fix it because of my lacking nix knowledge? maybe. i'll fix it later

ryoppippi: basically this is just a go binary, so i need to investigate

Mic92: There shouldn't really any nix knowledge required to use it correctly.

ryoppippi: yeah it's not about nix sorry, just am not cofident about it. thank you as always

PR #5131, opened 2026-05-22
pi: 0.75.3 -> 0.75.4

XYenon:
## Summary

- teach the npm updater to handle published `npm-shrinkwrap.json` files, falling back to generated `package-lock.json` when shrinkwrap entries lack integrity metadata for Nix prefetching
- remove upstream `npm-shrinkwrap.json` from the `pi` source wrapper so `buildNpmPackage` uses the generated lockfile
- update `pi` from 0.75.3 to 0.75.4

## Testing

- `nix shell --inputs-from .# nixpkgs#python3 --command python3 ./packages/pi/update.py`
- `nix build --accept-flake-config .#pi`
- `nix fmt`
- `python3 -m py_compile scripts/updater/npm.py packages/pi/update.py`

### DIFF
--- a/scripts/updater/npm.py
+++ b/scripts/updater/npm.py
@@ -1,11 +1,35 @@
 """NPM package utilities for Nix package updates."""
 
+import json
 import os
 import subprocess
 import tarfile
 import tempfile
 from pathlib import Path
+from typing import Any
 from urllib.request import urlretrieve
+
+
+def _can_prefetch_npm_lockfile(lockfile: Path) -> bool:
+    """Return whether the lockfile has enough metadata for Nix prefetching."""
+    data: dict[str, Any] = json.loads(lockfile.read_text())
+    packages = data.get("packages", {})
+    if not isinstance(packages, dict):
+        return True
+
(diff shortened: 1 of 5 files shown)


scripts/updater/npm.py
```diff
@@ -40,21 +64,35 @@ def extract_or_generate_lockfile(
...
         package_lock_src = package_dir / "package-lock.json"
+        shrinkwrap_src = package_dir / "npm-shrinkwrap.json"
 
         # Check if lockfile exists in tarball
         if package_lock_src.exists():
             output_path.write_text(package_lock_src.read_text())
             print("Updated package-lock.json from tarball")
             return True
 
+        # npm-shrinkwrap.json uses the same lockfile format and is published to
```
Mic92, line 75: do we need both lock files? To me it reads like shrinkwrap might be enough?

XYenon: For pi specifically, the published `npm-shrinkwrap.json` is missing integrity fields for a few tarball dependencies, so Nix's npm prefetcher cannot use it directly.

PR #5182, opened 2026-05-24
Add new package: reasonix at version 0.49.0

arch-fan:
## Summary

Adds a new package for Reasonix at version 0.49.0.
The update flow is set up to use nix-update with package-specific args so it follows the intended tag stream.

## Test plan

- Built the package locally:
  - nix build --accept-flake-config .#reasonix
- Verified repository checks:
  - nix flake check
- Verified update path:
  - nix-update works for this package with packages/reasonix/nix-update-args

- [X] `nix build .#<package>` succeeds
- [X] Package updates via `nix-update` or has a custom `update.py` if nix-update doesn't work

### DIFF
--- a/README.md
+++ b/README.md
@@ -360,6 +360,16 @@ Nix packages for AI coding agents and development tools. Automatically updated d
 - **Nix**: [packages/qwen-code/package.nix](packages/qwen-code/package.nix)
 
 </details>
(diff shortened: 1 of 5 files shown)


arch-fan:
great great. ty!

PR #5590, opened 2026-06-06
reasonix: fix Go package

arch-fan:
## Summary

Fix `reasonix` packaging:

- enable Go checks
- wrap the CLI with `ripgrep` and Linux `bubblewrap` for runtime search/sandbox support
- keep `nix-update` on plain `v*` tags
- idiomatic cleanup: use direct package inputs, string env values, and `passthru.category`

## Test plan

- [x] `nix build .#reasonix` succeeds

Also checked:

- `nix build --accept-flake-config .#checks.$(nix eval --raw --impure --expr builtins.currentSystem).pkgs-reasonix`
- `./result/bin/reasonix --version`
- `PATH= HOME=$(mktemp -d) ./result/bin/reasonix doctor --json`
- `nix run --accept-flake-config nixpkgs#nix-update -- --flake --version-regex '^v([0-9]+\.[0-9]+\.[0-9]+)$' reasonix`

> [!NOTE]

### DIFF
--- a/packages/reasonix/nix-update-args
+++ b/packages/reasonix/nix-update-args
@@ -1,3 +1,2 @@
---use-github-releases
 --version-regex
 ^v([0-9]+\.[0-9]+\.[0-9]+)$
--- a/packages/reasonix/nix-update-args
+++ b/packages/reasonix/nix-update-args
@@ -1,2 +1,3 @@
+--use-github-releases
 --version-regex
 ^v([0-9]+\.[0-9]+\.[0-9]+)$
--- a/packages/reasonix/package.nix
+++ b/packages/reasonix/package.nix
@@ -1,20 +1,25 @@
 {
-  pkgs,
   lib,
+  stdenv,
+  buildGoModule,
+  fetchFromGitHub,
   flake,
+  makeWrapper,
+  ripgrep,
+  bubblewrap,
   versionCheckHook,
   versionCheckHomeHook,
-  ...
 }:
+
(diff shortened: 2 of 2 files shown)


packages/reasonix/nix-update-args
```diff
@@ -1,3 +1,2 @@
---use-github-releases
```
Mic92, line 1: why was this dropped?

arch-fan: I dropped it because I tested nix-update with just the version regex and it correctly picked the latest release, so I thought the option was redundant

Mic92: Using the github api is usually better because it can make sure that all versions are resolved. The atom feed is limited. If packages do a lot of nightly updates than we might not see the last stable version. It requires a GITHUB_TOKEN to be set, however we always do have this in CI.

Mic92: Haha, I spoke too soon, we also have issues there. It just breaks after more versions :) https://github.com/numtide/llm-agents.nix/pull/5620

arch-fan: Ok, i just changed it hehe

arch-fan: Just read the latest comment 🤣🤣 should I remove it again?

Mic92: No i think it's still better. atm and it will be better in the next release.

Mic92: This is the fix: https://github.com/Mic92/nix-update/pull/613

PR #5832, opened 2026-06-13
bugfix: symlinking bwrap to non-wrapped codex

### LINKED ISSUE #5831: Codex issues with finding bwrap in path
### Package

codex

### Issue type

Runtime error

### Platform

x86_64-linux

### Nix sandbox enabled?

Yes

### Description

Hi,
I was having major issues with Codex recently and after some investigation I found that the provided Nix wrapper to codex did not properly handle the PATH environment variable to the 'main' codex binary. Resulting in the sandboxed commands failng making Codex pretty much unusable.

The Nix package wraps only $out/bin/codex to add bubblewrap to PATH:
```nix
wrapProgram $out/bin/codex \
--prefix PATH : ${lib.makeBinPath [ bubblewrap ]}
```

I think this is insufficient because Codex creates arg0 helper symlinks using std::env::current_exe().
In the Nix package, after wrapping, current_exe() resolves to the unwrapped binary `$out/bin/.codex-wrapped` instead of the wrapper script `$out/bin/codex`.

So Codex creates helpers like:
```
~/.codex/tmp/arg0/.../codex-linux-sandbox -> $out/bin/.codex-wrapped
```

When codex-linux-sandbox is invoked, it bypasses the wrapper script making the wrapper PATH containing bubblewrap not available to the sandbox.

This was causing errors like:
`bubblewrap is unavailable: no system bwrap was found on PATH`

## Proposed fix
Install/symlink Bubblewrap into Codex’s expected resource directory:

```nix
postFixup = lib.optionalString stdenv.hostPlatform.isLinux ''
mkdir -p $out/codex-resources
ln -s ${lib.getExe bubblewrap} $out/codex-resources/bwrap

wrapProgram $out/bin/codex \
  --prefix PATH : ${lib.makeBinPath [ bubblewrap ]}
'';
```

After rebuilding the package with that fix, on my end it works perfectly although I've seen times where the sandbox worked for some reason before the fix (no idea why).

### Error output / logs

```text

```

nestyy1337:
## Summary

The Codex package currently wraps only `$out/bin/codex` to add `bubblewrap` to `PATH`.
Codex creates arg0 helper symlinks using `std::env::current_exe()`, after wrapping it resolves to the unwrapped binary: `$out/bin/.codex-wrapped`
Helpers such as codex-linux-sandbox can point directly at .codex-wrapped and bypass the wrapper-provided PATH. When that happens sandboxed commands fail because bwrap is not visible to the sandbox launcher.
PR symlinks bubblewrap into Codex’s expected resource directory: `$out/codex-resources/bwrap`, Codex already checks this location as a bundled-resource fallback, so the sandbox helper can find bwrap even when invoked through the unwrapped binary.

## Test plan
- [x] `nix build .#codex` succeeds
- [x] Verified sandboxed commands work after rebuilding:
  - echo sandbox-test succeeds
  - writing to /tmp succeeds
  - writing outside the writable sandbox roots fails with Read-only file system

Closes #5831

### DIFF
--- a/packages/codex/package.nix
+++ b/packages/codex/package.nix
@@ -111,6 +111,9 @@ rustPlatform.buildRustPackage {
   '';
 
   postFixup = lib.optionalString stdenv.hostPlatform.isLinux ''
+    mkdir -p $out/codex-resources
+    ln -s ${lib.getExe bubblewrap} $out/codex-resources/bwrap
+
     wrapProgram $out/bin/codex \
       --prefix PATH : ${lib.makeBinPath [ bubblewrap ]}
   '';


packages/codex/package.nix
```diff
@@ -111,6 +111,9 @@ rustPlatform.buildRustPackage {
   '';
 
   postFixup = lib.optionalString stdenv.hostPlatform.isLinux ''
+    mkdir -p $out/codex-resources
+    ln -s ${lib.getExe bubblewrap} $out/codex-resources/bwrap
+
     wrapProgram $out/bin/codex \
       --prefix PATH : ${lib.makeBinPath [ bubblewrap ]}
```
Mic92, line 118: Do we still need this in this case?

PR #6498, opened 2026-07-06
claude-desktop: patch libraries with autoPatchelfHook

iainlane:
## Summary

The install phase currently patches every bundled ELF with a manual loop:

```sh
patchelf --set-interpreter … "$elf" 2>/dev/null || true
patchelf --set-rpath "$app_rpath" "$elf" 2>/dev/null || true
```

The redirected stderr and `|| true` mean a payload whose libraries are missing is patched into a broken state without failing the build; the missing library only surfaces when a user runs the affected component.

This switches to `autoPatchelfHook`, which discovers each payload's `DT_NEEDED` entries, resolves them against `buildInputs`, and **fails the build** when one can't be satisfied. Libraries reached only through `dlopen` (libGL, libsecret, the appindicator, PipeWire, Wayland) carry no `DT_NEEDED` entry, so they move to `runtimeDependencies` to stay on the RUNPATH. The wrapper keeps only the application directory on `LD_LIBRARY_PATH`, so the bundled GL/Vulkan libraries still find each other.

## This PR intentionally fails to build

Standing on its own, it surfaces a real gap rather than papering over it: the bundled `virtiofsd` (which backs Cowork's virtual machines) links `libseccomp.so.2` and `libcap-ng.so.0`, neither of which is in the closure today. The manual loop silently ships a `virtiofsd` that can't start; autoPatchelfHook makes it a build error:

```
auto-patchelf: 2 dependencies could not be satisfied
error: auto-patchelf could not satisfy dependency libseccomp.so.2 wanted by …/resources/virtiofsd
error: auto-patchelf could not satisfy dependency libcap-ng.so.0 wanted by …/resources/virtiofsd
```

#6497 adds those two libraries (plus `libXcursor`, which is `dlopen`ed and so is invisible to autoPatchelfHook — it has to be listed by hand either way). With that change in, this builds cleanly. The two PRs are independent; merging #6497 first (or folding it in here) makes CI green. I've kept them separate so this one's failure documents the gap it closes.

## Test plan

- `nix build .#claude-desktop` for `x86_64-linux` fails with exactly the two `virtiofsd` errors above, and nothing else.
- With #6497's libraries added on top, the same build succeeds; `virtiofsd`'s RUNPATH then resolves `libseccomp.so.2` and `libcap-ng.so.0`.
- `nix fmt` reports no changes.

- [ ] `nix build .#<package>` succeeds
- [x] Package updates via `nix-update` or has a custom `update.py` if nix-update doesn't work (unchanged; existing `update.py`)

### DIFF
--- a/packages/claude-desktop/package.nix
+++ b/packages/claude-desktop/package.nix
@@ -1,14 +1,15 @@
 {
   fetchurl,
   lib,
+  autoPatchelfHook,
   makeWrapper,
-  patchelf,
   stdenvNoCC,
   bintools,
   copyDesktopItems,
   makeDesktopItem,
 
-  # Linked dynamic libraries.
+  # Directly linked (DT_NEEDED); autoPatchelfHook resolves these from
+  # buildInputs and fails the build if any are missing.
   alsa-lib,
   at-spi2-atk,
   at-spi2-core,
@@ -21,7 +22,6 @@
   glib,
   gtk3,
   libdrm,
-  libglvnd,
   libX11,
   libxcb,
   libXcomposite,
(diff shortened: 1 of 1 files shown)


Mic92:
Superseded by #6532 (merged).

PR #6631, opened 2026-07-09
codex: also build the codex-code-mode-host binary

MrVampy:
## Problem

`codex` spawns a separate **`codex-code-mode-host`** binary for its command backend (shell execution + file edits). The package builds only `--package codex-cli`, so `codex-code-mode-host` is never built or installed. At runtime, no shell or file-edit command can start:

> The command backend broke … `codex-code-mode-host` is missing from its Nix store path, so no shell or file-edit command can start.

The `code-mode-host` crate (package name `codex-code-mode-host`) has been in the workspace since at least `rust-v0.142.5`; the command backend just didn't exercise it until recently, so the gap was latent.

## Fix

Add `--package codex-code-mode-host` to `cargoBuildFlags` so its binary is built and installed next to `codex` in `$out/bin` (codex resolves the host as a sibling of its own executable).

```diff
     cargoBuildFlags = [
       "--package"
       "codex-cli"
+      "--package"
+      "codex-code-mode-host"
     ];
```

I hit this on 0.144 and fixed it downstream by overriding `cargoBuildFlags` with the same addition; `codex-code-mode-host` then installs into `$out/bin` and the command backend works.

### DIFF
--- a/packages/codex/package.nix
+++ b/packages/codex/package.nix
@@ -75,6 +75,8 @@ rustPlatform.buildRustPackage (
     cargoBuildFlags = [
       "--package"
       "codex-cli"
+      "--package"
+      "codex-code-mode-host"
     ];
 
     nativeBuildInputs = [


calebrate:
Confirming this independently - we hit the same thing on `codex` 0.144.0 and arrived at exactly this diff. Adding a bit of detail that might be useful for whoever reviews.

**Resolution path.** Codex looks for the host binary as a sibling of its own executable, with `CODEX_CODE_MODE_HOST_PATH` as an override. The `wrapProgram` in `postFixup` doesn't interfere: `current_exe()` resolves to `$out/bin/.codex-wrapped`, so the sibling lookup still lands in `$out/bin`. Our error printed the full path, which matches:

```
ERROR codex_core::tools::router: error=failed to spawn code-mode host
/nix/store/…-codex-0.144.0/bin/codex-code-mode-host: No such file or directory (os error 2)
```

**What actually triggers it.** The failing models are the ones carrying `"tool_mode": "code_mode_only"` in `~/.codex/models_cache.json` — currently `gpt-5.6-sol`, `gpt-5.6-terra`, and `gpt-5.6-luna`. `gpt-5.5`, `gpt-5.4`, `gpt-5.4-mini`, and `gpt-5.3-codex-spark` have no `tool_mode` and are unaffected. This is consistent with your note that the crate has been in the workspace since `rust-v0.142.5` and the gap was latent: nothing exercised the host until these models shipped.

It also makes the bug easy to miss, since it only surfaces on the first tool call. A prompt that needs no tools succeeds on a `code_mode_only` model:

```bash
codex exec -m gpt-5.6-sol "reply with the single word OK"          # works
codex exec -m gpt-5.6-sol "run `echo hi` and report the output"    # fails to spawn the host
```

**Two small things that may save a reviewer time.** No `cargoHash` bump is needed — the vendor derivation already covers the whole workspace lock, and the change is confined to `cargoBuildFlags`. And for anyone who needs a stopgap before this lands, `CODEX_CODE_MODE_HOST_PATH` is genuinely honored: pointing it at a stub script got the host spawned (failing later at `code-mode host exited during handshake` rather than `ENOENT`), so a separately-built `codex-code-mode-host` can be wired in via the env var without rebuilding the ~412 MB `codex` binary and losing the `cache.numtide.com` hit.

Building it in `$out/bin` as this PR does is clearly the right default, though. Thanks for tracking it down — would be great to see this merged.

PR #6652, opened 2026-07-10
pi: add Bun runtime variant for aarch64-linux

parubets:
## Summary

Adds `pi-bun`, a Bun-runtime variant of the existing `pi` package.

The main motivation is Raspberry Pi / `aarch64-linux` support through Nixpkgs’ Bun runtime. Upstream Pi already ships a Bun-specific entrypoint at `dist/bun/cli.js` and builds standalone Bun binaries from that path, so this package reuses the existing npm tarball and runs that official Bun entrypoint with Nix’s `bun`.

## Implementation

- Extends `packages/pi/package.nix` with a small `runtime` parameter:
  - `runtime = "node"` keeps the existing `pi` behavior
  - `runtime = "bun"` builds the `pi-bun` variant
- Adds `packages/pi-bun` as a thin wrapper over `packages/pi`
- Keeps one source/hash/lock/update path via `packages/pi`
- Uses `bun.meta.platforms` for the Bun variant, which includes `aarch64-linux`
- Sets `PI_PACKAGE_DIR` in the wrapper, matching upstream’s documented Nix/Guix support path
- Runs the upstream Bun entrypoint directly:
  - `dist/bun/cli.js`
- Rewrites final executable JS shebangs from Node to Bun for the Bun variant, so the runtime closure does not pull in Node
- Adds `passthru.skipUpdate` support to updater discovery so `pi-bun` is not updated separately from `pi`

## Why a separate package?

`pi-bun` gives users an explicit package to select when they want the Bun runtime, while avoiding duplicated packaging state. The actual source, lockfile, dependency hash, and updater remain owned by `pi`.

This avoids maintaining two independent copies of the same upstream package while still exposing a clear user-facing runtime choice:

```bash
nix run .#pi
nix run .#pi-bun
```

## Validation

- `nix build --accept-flake-config .#pi`
- `nix build --accept-flake-config .#pi-bun`
- `nix build --accept-flake-config .#checks.$system.pkgs-pi`
- `nix build --accept-flake-config .#checks.$system.pkgs-pi-bun`
- `nix build --accept-flake-config .#checks.$system.meta-completeness`
- `nix build --accept-flake-config .#checks.$system.meta-maintainers`
- `nix build --accept-flake-config .#checks.$system.ast-grep`
- `nix build --accept-flake-config .#packages.aarch64-linux.pi-bun --dry-run`
- `nix fmt -- --ci`
- `./result-pi/bin/pi --version`
- `./result-pi-bun/bin/pi --version`
- `./result-pi-bun/bin/pi --help`
- Verified `pi-bun` runtime closure contains Bun and does not pull in Node
- Verified updater discovery selects `pi` and skips `pi-bun`
- Regenerated README package docs

### DIFF
--- a/.github/ci/discovery.py
+++ b/.github/ci/discovery.py
@@ -23,10 +23,11 @@ let
   pkgs = flake.packages.${config.system};
   isHidden = pkg: pkg.passthru.hideFromDocs or false;
   updateEvenIfHidden = pkg: pkg.passthru.updateEvenIfHidden or false;
-  shouldDiscover = pkg: !(isHidden pkg) || updateEvenIfHidden pkg;
+  skipUpdate = pkg: pkg.passthru.skipUpdate or false;
+  shouldDiscover = pkg: (!(isHidden pkg) || updateEvenIfHidden pkg) && !(skipUpdate pkg);
   getVersion = name:
-    if pkgs ? ${name} && pkgs.${name} ? version && shouldDiscover pkgs.${name}
-    then { inherit name; value = pkgs.${name}.version; }
+    if pkgs ? ${name} && pkgs.${name} ? version
+    then { inherit name; value = if shouldDiscover pkgs.${name} then pkgs.${name}.version else null; }
     else null;
 in
   if config.filter == null then
--- a/README.md
+++ b/README.md
@@ -321,6 +321,16 @@ Nix packages for AI coding agents and development tools. Automatically updated d
 
 </details>
 <details>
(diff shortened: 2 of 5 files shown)


Mic92:
I don't understand the motivation yet. What is the advantage of bun over node? I rather not want to have both bun and nodejs and would like to rather drop one over the other.

parubets:
The bun version load time is ~4 times faster. When used on small microvms for background agents tasks or with pi-subagents extension - this makes huge difference.
HMtNMqzbkAAlhvd

Mic92:
Fixed in main.

PR #6980, opened 2026-07-19
opencode2: init at 0.0.0-next-15824

iainlane:
## Summary

_Not sure if you generally take prerelease/beta packages. I wanted it for myself, figured I might as well submit here too._

OpenCode 2 is the preview of OpenCode's next-generation CLI. It combines the terminal interface and server in one executable, so it can run with a private server, reuse a background service, or connect to a remote server.

Unlike stable OpenCode, v2 is published through npm's next channel. Its top-level package selects a platform-specific Bun executable, so the GitHub release packaging used for v1 cannot be reused.

Package the four supported native executables, patch their Linux ELF dependencies, add ripgrep to PATH, and provide writable XDG directories for the version check. The package output contains only the executable and its wrapper, avoiding approximately 49 MiB of non-runtime source maps.

Extend the shared npm version fetcher with dist-tag support and use the same platform-binary updater as Copilot and Kilocode while tracking next. Regenerate the package documentation.

## Test plan

### DIFF
--- a/packages/opencode2/package.nix
+++ b/packages/opencode2/package.nix
@@ -1,36 +1,32 @@
 {
   lib,
   stdenv,
-  fetchurl,
   makeWrapper,
   wrapBuddy,
   ripgrep,
+  platformSource,
   versionCheckHook,
   versionCheckHomeHook,
 }:
 
 let
-  versionData = builtins.fromJSON (builtins.readFile ./hashes.json);
-  inherit (versionData) version hashes;
-
-  platformMap = {
-    x86_64-linux = "linux-x64";
-    aarch64-linux = "linux-arm64";
-    x86_64-darwin = "darwin-x64";
-    aarch64-darwin = "darwin-arm64";
+  # OpenCode 2 ships as platform-specific Bun executables on npm's next channel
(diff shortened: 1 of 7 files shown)


packages/opencode2/package.nix
```diff
@@ -0,0 +1,101 @@
...
+      OpenCode 2 is the preview of OpenCode's next-generation CLI. The single
+      executable includes the terminal interface and server, and can run with
+      a private server, reuse a background service, or connect to a remote
+      server.
+    '';
+    homepage = "https://opencode.ai";
+    changelog = "https://github.com/anomalyco/opencode/commits/v2";
+    downloadPage = "https://www.npmjs.com/package/@opencode-ai/cli?activeTab=versions";
+    license = lib.licenses.mit;
+    sourceProvenance = with lib.sourceTypes; [ binaryNativeCode ];
```
Mic92, line 77: Error: New package 'opencode2' has empty meta.maintainers. Add yourself (see lib/default.nix for the local maintainer list).
Error: Process completed with exit code 1.

iainlane: Did it, cheers

PR #7251, opened 2026-07-27
hermes-agent: expose modules to slash workers

mulatta:
## Summary

Dashboard slash commands fail under Nix because slash workers re-exec the bare Python interpreter, which cannot import either Hermes application modules or packaged dependencies. Prefix the Hermes wrapper's `PYTHONPATH` with both site-packages roots so inherited slash-worker environments can import `tui_gateway` and dependencies such as `yaml`, and add an install check covering that bare-interpreter path.

This is a packaging workaround until the upstream slash-worker callsite honors `HERMES_PYTHON` and `HERMES_PYTHON_SRC_ROOT`; related upstream work is tracked in NousResearch/hermes-agent#51277.

## Test plan

Verified `nix build .#hermes-agent` succeeds. Deployed the resulting package and confirmed the Dashboard's `/tools list` command returns the built-in toolsets instead of `error: slash worker exited`.

- [x] `nix build .#hermes-agent` succeeds

> [!NOTE]

### DIFF
--- a/packages/hermes-agent/package.nix
+++ b/packages/hermes-agent/package.nix
@@ -373,6 +373,12 @@ python3.pkgs.buildPythonApplication {
     "--set"
     "HERMES_PYTHON_SRC_ROOT"
     "${placeholder "out"}/${python3.sitePackages}"
+    # Dashboard slash workers re-exec the bare sys.executable, bypassing both
+    # the application wrapper and the dependency environment.
+    "--prefix"
+    "PYTHONPATH"
+    ":"
+    "${placeholder "out"}/${python3.sitePackages}:${pythonEnv}/${python3.sitePackages}"
     "--set"
     "HERMES_NODE"
     "${nodejs}/bin/node"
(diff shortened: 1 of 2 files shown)


packages/hermes-agent/package.nix
```diff
@@ -327,6 +327,12 @@ python3.pkgs.buildPythonApplication {
     "--set"
     "HERMES_PYTHON_SRC_ROOT"
     "${placeholder "out"}/${python3.sitePackages}"
+    # Dashboard slash workers re-exec the bare sys.executable, bypassing both
+    # the application wrapper and the dependency environment.
+    "--prefix"
+    "PYTHONPATH"
+    ":"
+    "${placeholder "out"}/${python3.sitePackages}:${pythonEnv}/${python3.sitePackages}"
```
Mic92: Will this not also leak into any bash tool call?

Mic92: Maybe better to have sys.path patched in code

Mic92:
Pushed an alternative on top of your commit: instead of prefixing PYTHONPATH on the wrapper (which leaks Hermes' Python deps into every subprocess it spawns, including Python run inside the user's project), the slash-worker spawn in `tui_gateway/server.py` is patched to use `HERMES_PYTHON` and to pass `HERMES_PYTHON_SRC_ROOT` via PYTHONPATH only for that subprocess (`slash-worker-hermes-python.patch`). The install check now probes exactly that path. Verified: build succeeds and `PYTHONPATH=$out/site-packages $pythonEnv/bin/python3 -c 'import tui_gateway.slash_worker, yaml'` works. The patch can be dropped once upstream #51277 lands.

mulatta:
Rebased the downstream patch for Hermes Agent v2026.7.30. The slash-worker environment setup changed to use `build_subprocess_env`, so the previous patch no longer applied. Updated the patch for the new callsite and confirmed the Nix package builds successfully.

PR #7759, opened 2026-08-09
qoder-cli-cn: init at 1.1.17

RyougiShiki-214:
## Summary

Add `qoder-cli-cn`, packaging the China-channel build of the Qoder CLI. It is a
bun-compiled binary distributed through the official `static.qoder.com.cn`
manifest, and is modeled directly on the existing `qoder-cli` package.

Differences from `qoder-cli`:

- Manifest/download source: `https://static.qoder.com.cn/qoder-cli-cn/channels/manifest.json`
- Binary name: `qoderclicn`
- Metadata points at `https://qoder.cn`

Like `qoder-cli`, it uses `wrapBuddy` on Linux and `dontStrip` (bun runtime), and
ships a custom `update.py` because nix-update cannot handle the custom manifest
source.

## Test plan

- [x] `nix build .#qoder-cli-cn` succeeds
- [x] Package updates via a custom `update.py` (nix-update doesn't work for the manifest source)

Additional verification run locally:

- `./result/bin/qoderclicn --version` prints `1.1.17`
- `nix flake check` — all checks passed
- `nix fmt` applied
- README regenerated via `./scripts/generate-package-docs.py`
- `check_maintainers.py` passes (new package declares a maintainer)

### DIFF
--- a/packages/qoder-cli-cn/package.nix
+++ b/packages/qoder-cli-cn/package.nix
@@ -47,7 +47,7 @@ stdenv.mkDerivation {
   passthru.category = "AI Coding Agents";
 
   meta = with lib; {
-    description = "终端原生的 AI 编程搭档，也是可被集成的智能体引擎。";
+    description = "Qoder CLI (mainland China edition) - terminal-based AI coding assistant for China-region accounts";
     homepage = "https://qoder.cn";
     changelog = "https://qoder.cn/changelog";
     downloadPage = "https://qoder.cn/download";
--- a/packages/qoder-cli-cn/package.nix
+++ b/packages/qoder-cli-cn/package.nix
@@ -1,0 +1,64 @@
+{
+  lib,
+  flake,
+  stdenv,
+  fetchurl,
+  wrapBuddy,
+  versionCheckHook,
+  versionCheckHomeHook,
+}:
+
+let
+  versionData = builtins.fromJSON (builtins.readFile ./hashes.json);
+  inherit (versionData) version platforms;
+
+  platform = stdenv.hostPlatform.system;
+  src = platforms.${platform} or (throw "Unsupported system: ${platform}");
+in
+stdenv.mkDerivation {
+  pname = "qoder-cli-cn";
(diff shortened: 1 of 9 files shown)


Mic92:
> Differences from `qoder-cli`:
> 
>     * Manifest/download source: `https://static.qoder.com.cn/qoder-cli-cn/channels/manifest.json`
> 
>     * Binary name: `qoderclicn`
> 
>     * Metadata points at `https://qoder.cn`

I don't get the difference. What are the functional differences beyond where binaries are coming from?

packages/qoder-cli-cn/package.nix
```diff
@@ -0,0 +1,64 @@
...
+
+  nativeInstallCheckInputs = [
+    versionCheckHook
+    versionCheckHomeHook
+  ];
+
+  passthru.category = "AI Coding Agents";
+
+  meta = with lib; {
+    description = "终端原生的 AI 编程搭档，也是可被集成的智能体引擎。";
```
Mic92: Is this not meant to be used outside of China?

RyougiShiki-214: Sure, sorry — my description wasn't very clear!

Alibaba ships Qoder as two separate builds: an international edition (qoder.com, binary qodercli) and a mainland China edition (qoder.cn, binary qoderclicn, this package). Same versions, but different artifacts — not mirrors.

The split is because of region-locked accounts/subscriptions: China accounts (billed via Chinese payment channels) can only sign in to the CN build, while international accounts only work with the international one. So users in mainland China simply can't use qoder-cli — the CN build is their only option, hence a separate package.

Happy to reword the description to make this clearer!

Mic92: Okay, do you mind changing the description of the package to English to make it to reduce the confusion of what this is about?

RyougiShiki-214: Done — reworded the description in English, thanks for the suggestion!

PR #7888, opened 2026-08-12
treefmt: quote the attribute after inputs.

zimbatm:
## Description

New law: the attribute after `inputs.` must be quoted. :stuck_out_tongue:

- `scripts/quote-flake-inputs.py` — a treefmt formatter that rewrites `inputs.foo` to `inputs."foo"`. It is a minimal Nix lexer, not a regex, so comments, `"…"`/`''…''` string contents, and `inputs.${...}` interpolations are untouched. Idempotent.
- Registered in the `nix` pipeline at priority 0, before deadnix and nixfmt. nixfmt preserves quoted attr names, so the two do not fight.
- Applied to the tree: `flake.nix`, `lib/default.nix`, and the `flake.inputs.bun2nix` references in package wrappers (13 files).

## Testing

- Edge-case file covering already-quoted names, interpolation, comments, block comments, both string kinds, and `myinputs.foo` boundary — all handled; second run is a no-op.
- `nix fmt` clean (ruff, mypy, nixfmt all pass).
- `nix build .#formatter` succeeds; affected packages (`bun2nix`, `backlog-md`) still evaluate.

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -11,30 +11,30 @@
     systems.url = "github:nix-systems/default";
     blueprint = {
       url = "github:numtide/blueprint";
-      inputs.nixpkgs.follows = "nixpkgs";
-      inputs.systems.follows = "systems";
+      inputs."nixpkgs".follows = "nixpkgs";
+      inputs."systems".follows = "systems";
     };
     treefmt-nix = {
       url = "github:numtide/treefmt-nix";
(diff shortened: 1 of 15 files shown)


zimbatm:
Superseded — main now has ast-grep-based lint rules (#7886).

PR #7890, opened 2026-08-12
mcptoon: drop obsolete default.nix

zimbatm:
## Description

Leftover from #7839 racing the package-scope restructure: the trivial `default.nix` wrapper is dead code now — the scope calls `package.nix` directly.

## Testing

`nix build .#mcptoon` produces the same store path as before the removal (cache hit), and the CLI runs.

### DIFF
--- a/packages/mcptoon/default.nix
+++ b/packages/mcptoon/default.nix
@@ -1,1 +1,0 @@
-{ pkgs, ... }: pkgs.callPackage ./package.nix { }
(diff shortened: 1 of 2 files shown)


zimbatm:
Also fixes the version mismatch: upstream tagged v0.2.2 without bumping `__version__` (the CLI banner printed "v0.2.1"). `postPatch` now aligns it with the tag — `mcptoon` banner reports v0.2.2 after the rebuild.

PR #7908, opened 2026-08-12
ci: isolate updater code from the App token, confine its diff

zimbatm:
## Description

The update workflow ran per-package `update.py` scripts (63 of them) and `nix-update` with the GitHub App token (`contents`+`pull-requests` write) exported in their environment, in a checkout whose git config held push credentials, and later committed the entire work tree with `git add .`. Merged updater code could therefore exfiltrate the token or smuggle arbitrary changes into an auto-merged PR.

This closes the channels:

**Token isolation** (`update-flake.yml`)
- The App token is minted *after* the "Perform update" step, so it does not exist while updater code runs. The updater gets the default workflow token (job `permissions: contents: read`) purely for GitHub API rate limits — the only thing updater code uses it for (`scripts/updater/http.py`).
- `actions/checkout` uses `persist-credentials: false`; the PR-creation step sets an authenticated remote URL explicitly. Same treatment for the `update-readme` job.

**Step poisoning** — `.github/ci` is copied to `$RUNNER_TEMP` before the updater runs and `update.py`/`create_pr.py` execute from there, so the updater cannot rewrite the script that later runs with the token.

**Diff confinement** (`create_pr.py`) — before committing, the changed paths are checked against the update's territory: `packages/<name>/` for a package update, `flake.lock` for a flake input. Anything else fails the job before commit/push. Renames/copies out of the package dir are treated as violations (both sides of the rename are checked). Maintainer fixup commits already on a reused `update/*` branch are unaffected — only the work tree the updater just dirtied is checked.

**Injection hygiene** — `matrix.*` and updater-controlled outputs (`new_version`) now reach shell via `env:` instead of inline `${{ }}` interpolation; a crafted version string can no longer inject shell into the token-bearing step.

Residual risk (by design): an updater can still shape the version/hash contents of its own package's update PR; that diff is public and gated by CI + the ast-grep rules before auto-merge.

## Testing

- Confinement unit-tested in a scratch repo: in-dir change accepted; stray file, wrong-territory change, and rename-out-of-dir all rejected with exit 1.
- Grepped all 63 `packages/*/update.py` + `scripts/updater/`: none legitimately writes outside its package dir, so no updater trips the new check.
- `actionlint` reports the same 12 pre-existing findings on main and this branch (stale action schema + summary-step style nits); nothing new introduced.
- `nix fmt` clean (ruff, mypy, yamlfmt).

### DIFF
--- a/.github/ci/create_pr.py
+++ b/.github/ci/create_pr.py
@@ -127,32 +127,16 @@ def allowed_roots(update_type: UpdateType, name: str) -> tuple[str, ...]:
 
 
 def changed_paths(repo: Path) -> list[str]:
-    """List every path that differs from origin/main.
+    """List every work-tree path that differs from origin/main.
 
-    Covers staged, unstaged, untracked, and already-committed-on-branch
-    changes — a malicious updater could hide changes behind its own commit.
+    Only the work tree matters: publishing copies files from it, so the
+    index and branch commits are irrelevant. Tracked changes (including
+    ones a malicious updater committed) show up in the diff against
+    origin/main; new files via ls-files.
     """
-    paths: set[str] = set()
-
-    out = git_ro(repo, "status", "--porcelain=v1", "-z")
-    tokens = out.split("\0")
-    i = 0
-    while i < len(tokens):
-        entry = tokens[i]
-        i += 1
-        if not entry:
-            continue
-        status, path = entry[:2], entry[3:]
-        paths.add(path)
-        # renames/copies carry the source path in the next token
-        if "R" in status or "C" in status:
-            paths.add(tokens[i])
-            i += 1
-
-    out = git_ro(repo, "diff", "--name-only", "-z", "origin/main")
-    paths.update(p for p in out.split("\0") if p)
-
-    return sorted(paths)
+    tracked = git_ro(repo, "diff", "--name-only", "-z", "origin/main")
+    untracked = git_ro(repo, "ls-files", "--others", "--exclude-standard", "-z")
+    return sorted({p for p in (tracked + untracked).split("\0") if p})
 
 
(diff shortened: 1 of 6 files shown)


zimbatm:
Pushed a second commit closing two gaps @zimbatm flagged in review discussion:

**The confinement check ran against state the updater controlled.** Even with `create_pr.py` stashed out of the work tree, it still committed/pushed from the updater's checkout — whose `.git` the updater could poison (hooks, `core.fsmonitor`, `url.*.insteadOf` redirecting the token-bearing push). The stashed *script* was intact but operated on rigged *state*.

Two layers now:

1. **bubblewrap sandbox** around `update.py`/`nix-update` (Linux). Filesystem read-only except the package's own dir, `/nix`, and a fresh tmpfs `HOME`+`/tmp`; network stays up for upstream APIs. Verified: escape writes to `.git/hooks`, sibling package dirs, `/etc`, and repo root are all discarded/blocked, while writes to the package's own dir persist. Falls back to unsandboxed-with-warning where user namespaces are disabled (checks, and lifts the Ubuntu 24.04 AppArmor restriction via the runner's passwordless sudo); `UPDATE_SANDBOX=0` opts out locally.
2. **Publish from a pristine clone.** `create_pr.py` shallow-clones `main` fresh, copies only the allowed paths across, and commits/pushes from there — the App token never enters a repo the updater touched. `gh` runs from the clone; read-only git probes on the dirty tree disable fsmonitor/hooks. Confinement now also diffs `origin/main` so a stray hidden behind the updater's own commit is still caught.

All updaters checked for compatibility: none writes outside its own package dir (the `happy-cli`/`coding-agent` refs are a git-log path and a tarball prefix), and `scripts/updater/` imports stay readable via the read-only root bind.

PR #7933, opened 2026-08-13
scripts: port generate-package-docs to Nushell (pilot)

zimbatm:
## What

A concrete Nushell-vs-Python pilot, per the discussion about replacing the repo's Python. I picked the single best candidate to judge Nushell on: `generate-package-docs.py` — 188 self-contained lines, no shared-library dependency, no security surface, and output that can be diffed byte-for-byte.

- **`generate-package-docs.nu` replaces the `.py`.** The metadata heavy-lifting stays in `generate-package-docs.nix` (unchanged). Output is **byte-for-byte identical** to the Python generator — verified by wiping the generated README section, regenerating with the Nushell script, and diffing against the Python baseline (`diff` clean).
- **Self-bootstraps Nushell via a nix shebang** (`#!/usr/bin/env nix` → `nix shell --inputs-from .# nixpkgs#nushell`), the same convention the package updaters use. CI needs no `nu` on the runner; the invocation path is unchanged apart from the extension.
- **`devshell` gains `pkgs.nushell`**; **treefmt gains a `nu-check` parse check.** Nushell has no mature autoformatter yet (nufmt is alpha), so this is check-only — consistent with the existing `mypy-check`/`ruff-check`/`ast-grep-check` steps. Verified it fails on a malformed `.nu`.
- Updated the three references: `AGENTS.md`, `check-readme.yml`, `update-flake.yml`.

## Notes for evaluation

What the port teaches about Nushell for this codebase:

- **Wins:** the data pipeline is clean — `nix eval --json | from json | transpose | where | sort-by` reads better than the Python dict/loop equivalent, and native JSON + `http get` would shine in the updater library.
- **Sharp edges hit:** literal `(` in an interpolated string opens a subexpression (broke the markdown link targets until escaped `\(`); `nu-check` prints `true`/`false` but always exits 0 (needs a wrapper); no static type checking (the shared updater lib currently leans on mypy).

**Deliberately not ported:** the 65 contributor-authored `packages/*/update.py`. That's where switching languages would raise the contribution barrier most, and where the mypy/security-control arguments against a wholesale move are strongest. This pilot is meant to inform that larger decision, not prejudge it.

## Testing

- Byte-for-byte README parity vs the Python generator (perturb → regenerate → `diff` clean).
- "No changes" idempotent path and direct `./scripts/generate-package-docs.nu` shebang invocation both work.
- `nix fmt` clean; `nu-check` catches a broken `.nu`; treefmt closure builds.


Mic92:
Does nushell has type checking?

Mic92:
I think for the updater scripts, python is better because it's more of a mature programming language and the typechecking actually helps when refactoring.

PR #8029, opened 2026-08-14
[experiment] corepkgs: a nixpkgs-free packaging system

zimbatm:
## corepkgs: a nixpkgs-free packaging system

Introduces **corepkgs**, a self-contained packaging system that builds the repo's agent tools with **no nixpkgs and no stdenv** at build time. The root `llm-agents.nix` flake becomes pure *agent output*; all build machinery lives in `corepkgs/` as a standalone sub-flake.

## Why

Look at what this repo actually packages: a large, fast-moving set of **prebuilt-binary** AI CLIs. For most of them the "build" is almost entirely *fetch a release artifact → make it run on NixOS (rewrite the ELF interpreter/rpath, or loader-wrap a `bun --compile` binary) → wrap it*. That is a tiny sliver of what `stdenv` exists to do. corepkgs does exactly that sliver with a static busybox + nushell seed and nothing else.

Two motivations, one already realized and one directional:

- **A small, source-bootstrappable trusted base (now).** `stdenv` is a large, opaque thing to trust and build for "download a binary and patch its rpath." The corepkgs build is driven by a static busybox + a static nushell — a base small enough to audit, and small enough to eventually **bootstrap from source** (our own release tarballs, later GNU Mes) so the whole chain is trustworthy end to end. This is true today.
- **A path to zero-nixpkgs eval + full substitutability (directional).** When a package (and its pins) is fully nixpkgs-free, eval is dramatically cheaper and outputs stop tracking the consumer's nixpkgs at all.

**Measured, honestly** (x86_64-linux, cold eval, this branch):

| | eval a single package | full package-set eval |
|---|---|---|
| standalone corepkgs (zero nixpkgs, storePath pins) | **59 ms** | — |
| the repo flake (nixpkgs imported) | ~1.2–1.7 s | ~19 s |

So the fully nixpkgs-free path is **~20× faster** to eval. But note the honest caveat: **inside the current repo flake there is no eval win yet** — full-set eval is ~19 s either way, and a lone corepkgs package is even a touch *slower* because it forces the formatelf pin. The flake still imports nixpkgs (for the pins and the ~130 un-ported packages), and flake **pure eval forbids `builtins.storePath`**, so the pins can't yet be the fast storePath ones. The eval-speed and store-path-decoupling wins are therefore **latent**: they materialize as more of the tree becomes corepkgs and the pins move to storePath. What corepkgs buys *today* is the smaller trusted build base and the architecture to get there; nixpkgs stays for the genuinely source-built packages.

## How it works

- **Truly-static seed**: a static busybox + a truly-static musl **nushell** are the entire bootstrap. nushell is the build-script runtime — `__structuredAttrs` exposes the derivation attrs as JSON that nushell `open`s natively, so params pass as real lists/records, not string-munged env vars.
- **No stdenv**: `mk/naked.nix` builds a derivation from the nushell seed directly (~10 lines vs stdenv's ~2000). `mk/binary.nix` is the nixpkgs-free equivalent of `platformSource + autoPatchelf + makeWrapper`: fetch a release artifact, unpack (`unpack = "auto"` infers zip vs tar per platform), patch (or loader-wrap bun `--compile` binaries), wrap.
- **ELF patching** via [formatelf](https://github.com/Mic92/formatelf) (a patchelf replacement); **Mach-O** needs none — darwin CLIs link the always-present system `libSystem` via dyld.
- **Shared source of truth**: corepkgs builds reuse each package's existing `hashes.json` (what nix-update bumps) + an interpolated URL template — no duplicated, drift-prone hashes.
- **Sub-flake**: `corepkgs/` is importable (`core = import ./corepkgs { system; pkgs; }` → `core.lib.mkBinary`) and a real standalone flake (`nix build ./corepkgs#formatelf`). A `package.nix` that declares `mkBinary` *is* a corepkgs build; `callPackage` resolves it from the scope, and it still carries `meta` + `passthru.category` + `passthru.updater` so meta-completeness, the README generator, and the declarative updater treat it like any package.

## Status

- ✅ Seed → `mk/*` constructors → toolchains (bun/node/rust/zig/python) → `mkBinary` → FHS checks.
- ✅ **Darwin + multi-arch**: one `package.nix` serves x86_64/aarch64 linux and aarch64-darwin.
- ✅ **23 prebuilt-binary agents swapped** to corepkgs outright (nixpkgs build removed), each building **and running**; per-package FHS checks assert every output is store-only.
- ✅ corepkgs is a clean, self-contained **sub-flake** (`mk/` constructors, owned primitives, `core.lib` API).
- 🚧 Follow-ups: move the formatelf autoPatchelf hook fully into corepkgs (root purely agents); generalize `mk/cargo.nix` for source-built rust; corepkgs wrap-buddy support (swamp); a nested-`hashes.json` reader (qoder), darwin JRE dir-install (junie).

> Bootstrap note: the seed is currently trusted static upstream builds (busybox, nushell). The next steps toward a real trust chain are our own source-built bootstrap tarballs on a GitHub release, and eventually a GNU Mes bootstrap; the static builds are the pragmatic starting point, and the seed layer is kept small and swappable for exactly this.


Mic92:
> Toolchains from upstream prebuilt binaries: bun 1.3.14, node v22.14.0, rust 1.90.0, zig 0.16.0. zig cc links fully static rust executables — no glibc, no nixpkgs.

I don't think I want to be maintaining all  those toolchains in this project myself or use upstream binaries that are patched to work well in the nix context. Especially on the macOS side there are certain patches applied to make it properly work against the nixpkgs provided framework, which has be proven to be more robust to w.r.t to macOS upgrades compared to the homebrew approach.

Mic92:
> Toolchains from upstream prebuilt binaries: bun 1.3.14, node v22.14.0, rust 1.90.0, zig 0.16.0. zig cc links fully static rust executables — no glibc, no nixpkgs.

there is a quiet a bit of integration work that comes from upstream, that I wouldn't want to miss or maintain.

markasoftware:
but like, why? almost everyone using llm-agents.nix is going to have other nixpkgs dependencies, and some of the packages in here do in fact do heavy builds involving lots of the parts of stdenv

PR #8177, opened 2026-08-19
t3code: split provider wrapper from unwrapped package

merrkry:
## Summary

- Move the heavyweight t3code source build into a private `unwrapped.nix` derivation.
- Make `t3code` a lightweight wrapper that adds `providerPackages` to `PATH`.
- Allow users to override the default provider list with `t3code.override { providerPackages = [ ... ]; }`.
- Preserve the existing public interfaces: `t3code` provides `t3`, while `t3code-desktop` provides the desktop application.
- Point `nix-update` at `unwrapped.nix` and extend the packaging checks for wrapper overrides.

This decouples provider updates from the roughly 1 GiB t3code build. Updating packages such as `claude-code`, `codex`, or `opencode` now only rebuilds the small wrapper instead of downloading and rebuilding t3code itself.

## Test plan

- [x] `nix fmt`
- [x] `nix flake check --accept-flake-config --no-build path:.`
- [x] `nix build --accept-flake-config path:.#t3code path:.#t3code-desktop`
- [x] `nix build --accept-flake-config path:.#checks.x86_64-linux.t3code-packaging`
- [x] Verified that overriding `providerPackages` reuses the same unwrapped derivation and only changes the wrapper derivations.
- [x] Temporarily downgraded t3code from 0.0.33 to 0.0.32 and verified that the updater restored 0.0.33 with the correct source, Cargo, and pnpm dependency hashes.

## Discloure

This PR is assisted by OpenAI Codex with manual smoke tests.

### DIFF
--- a/checks/t3code-packaging.nix
+++ b/checks/t3code-packaging.nix
@@ -9,21 +9,27 @@ let
   packages = flake.packages.${system};
   t3code = packages.t3code;
   desktop = packages.t3code-desktop;
+  t3codeWithoutProviders = t3code.override { providerPackages = [ ]; };
 in
 assert t3code.meta.mainProgram == "t3";
 assert desktop.meta.mainProgram == "t3code-desktop";
 assert !(desktop ? version);
 assert t3code.passthru ? resourceMonitor;
+assert t3code.unwrapped ? override;
+assert t3code.pnpmDeps == t3code.unwrapped.pnpmDeps;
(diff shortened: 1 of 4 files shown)


merrkry:
Note: to override `t3code-desktop`, you need to write like

```nix
(llm-agents.t3code.override {
  providerPackages = with llm-agents; [
    codex
  ];
}).desktop
```

instead of `t3code-desktop.override {}`.

Not sure if there is a way to make this more intuitive...

PR #8340, opened 2026-08-23
chatgpt: bundle primary runtime dependencies

abkein:
### Problem

This is primarily related (at least the way I faced it) to Codex Security, both tab and plugin.

First, `codex-security` plugin ships with its MCP, which fails to start with message:
```
MCP client for `codex-security` failed to start: MCP startup failed: handshaking with MCP server failed: connection closed: initialize response
```
Logs (usually in `$CODEX_HOME/logs_2.sqlite`) reveal the true cause:
```
MCP server stderr (./scripts/launch_codex_security_mcp): Could not start dynamically linked executable: <$XDG_CACHE_HOME>/codex-runtimes/codex-primary-runtime/dependencies/node/bin/node NixOS cannot run dynamically linked executables intended for generic linux environments out of the box.
```

Secondly, the same problem appears, when one opens Security tab in Codex (in the ChatGPT desktop app):
```
Could not start dynamically linked executable: <$XDG_CACHE_HOME>/codex-runtimes/codex-primary-runtime/dependencies/python/bin/python3 NixOS cannot run dynamically linked executables intended for generic linux environments out of the box.
```

### Things done

It seems a good idea to simply bundle `codex-primary-runtime`. The downloader is `usr/lib/chatgpt/resources/app.asar`. It fetches: `https://persistent.oaistatic.com/codex-primary-runtime/latest/linux-x64/LATEST.json`.

`codex-security` already supports the solution cleanly:
- `plugins/cache/openai-curated-remote/codex-security/0.1.21/.mcp.json` forwards PYTHON.
- Its bundled server resolves `process.env.PYTHON` before falling back to `~/.cache/codex-runtimes/.../python3`.

No `app.asar` patch is needed for this particular failure. The package structure is:
1. Add a `codex-primary-runtime-python` derivation.
2. Extend `packages/chatgpt/update.py` and `packages/chatgpt/hashes.json` with the runtime version, URL, and hash.
3. Extract only `codex-primary-runtime/dependencies/python`.
4. Patch it with `wrapBuddy`, not `formatelf`, because `artifact_tool_v2` contains a Bun-compiled daemon.
5. Wrap ChatGPT with:
```sh
--set PYTHON ${codex-primary-runtime-python}/bin/python3
--set CODEX_MCP_NODE_PATH ${nodejs}/bin/node
```
The upstream Python archive also has malformed absolute symlinks into its build-time `/tmp` directory and an unused malformed `libpython3.so` ABI stub. Those need deterministic cleanup during installation.

 I verified:
- Python 3.12.13 starts normally.
- NumPy, pandas, lxml, Pillow, cryptography, pydantic, PDFium, ReportLab, Tkinter, and artifact_tool_v2 import successfully.
- The proprietary artifact RPC daemon starts and creates its Unix socket.
- The Bun payload tail remains byte-for-byte identical after wrapBuddy.
- AArch64 package evaluation; it remains without the Python override because OpenAI currently publishes no Linux ARM64 runtime manifest.

One limitation: this fixes `codex-security` and other components honoring `PYTHON`. ChatGPT’s general `load_workspace_dependencies` mechanism will still advertise the downloaded cache runtime, including its other generic Linux binaries. Packaging and redirecting the entire runtime would be a broader follow-up.

## Test plan

> [!NOTE]

### DIFF
--- a/checks/chatgpt-packaging.nix
+++ b/checks/chatgpt-packaging.nix
@@ -1,0 +1,35 @@
+{
+  pkgs,
+  flake,
+  system,
+}:
+
+let
+  chatgpt = flake.packages.${system}.chatgpt or null;
+  chatgptWithPrimaryRuntime =
+    if chatgpt != null then
+      chatgpt.override {
+        withPrimaryRuntime = true;
+      }
+    else
+      null;
+in
+assert
+  chatgpt == null
+  || (
+    !chatgpt.withPrimaryRuntime
+    && chatgpt.primaryRuntime == null
+    && chatgpt.runtimePython == null
+    && (
+      system != "x86_64-linux"
+      || (
+        chatgptWithPrimaryRuntime.withPrimaryRuntime
+        && chatgptWithPrimaryRuntime.primaryRuntime != null
+        && chatgptWithPrimaryRuntime.runtimePython != null
+        && chatgptWithPrimaryRuntime.drvPath != chatgpt.drvPath
+      )
+    )
+  );
+pkgs.runCommand "chatgpt-packaging-check" { } ''
+  touch $out
+''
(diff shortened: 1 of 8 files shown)


lukts30:
There is another NixOS runtime issue: the bundled Codex cannot execute local commands because ChatGPT provides neither `bwrap` on PATH nor codex-resources/bwrap. 

The standalone Codex package already handles this. Could this PR also add bubblewrap to the ChatGPT wrapper via sth like  `--prefix PATH : ${lib.makeBinPath [ bubblewrap ]}`?

Mic92:
merged the other pr.

PR #8396, opened 2026-08-25
chatgpt: patch downloaded codex runtime for NixOS at launch

Mic92:
Alternative to #8340.

The ChatGPT app downloads a generic-linux `codex-primary-runtime` (Python 3.12 + site-packages, Node, LibreOffice headless, poppler, libheif, git) into `~/.cache/codex-runtimes` and execs binaries from it. On NixOS these fail to start.

Rather than vendoring the ~1 GB proprietary bundle as a derivation and patching four more minified functions in `app.asar`, the wrapper now sources a small script that runs `auto-formatelf` (our autoPatchelf replacement) over the cache directory. It rewrites interpreter/RPATH in place against glibc, libstdc++, zlib, libxcrypt-legacy, curl, nss/nspr. A stamp file keyed on `runtime.json` + the store paths makes it a ~10 ms no-op after the first ~1 s run, and re-triggers when OpenAI ships a new runtime or nixpkgs bumps. A short-lived background loop catches the initial download during the first session.

Tested against runtime 26.819.11345: python (numpy, lxml, pydantic-core), node, soffice, pdftoppm, heif-convert and git-credential-manager all run. Known gap: the bundled *fallback* git's HTTP helpers want Debian's `libcurl-gnutls.so.4` (`CURL_GNUTLS_3`), which nixpkgs does not provide; the app prefers git from `PATH` anyway.

No asar changes, nothing redistributed, survives minifier renames.

Also exposes `formatelf.bin` for use outside build hooks.

### DIFF
--- a/packages/chatgpt/patch-runtime.sh
+++ b/packages/chatgpt/patch-runtime.sh
@@ -1,0 +1,44 @@
+# shellcheck shell=bash
+# Sourced by the chatgpt wrapper on Linux. The app downloads a generic-linux
+# "codex-primary-runtime" (python, node, git, libreoffice, ...) into
+# ~/.cache/codex-runtimes and execs binaries from it. Rewrite their ELF
+# interpreter/RPATH in place whenever the bundle changes so they run on NixOS.
+
+chatgpt_patch_runtimes() {
+  local root="$HOME/.cache/codex-runtimes" dir stamp want interp
+  interp=$(<@dynamicLinker@)
+  for dir in "$root"/*/; do
+    [[ -f "$dir/runtime.json" ]] || continue
+    # Re-patch when either the bundle or our toolchain (store paths) changed.
+    want=$(
+      cat "$dir/runtime.json"
+      echo "$interp @libs@"
+    )
+    stamp="$dir/.nix-patched"
+    [[ -f $stamp && "$(<"$stamp")" == "$want" ]] && continue
+    echo "chatgpt: patching $dir for NixOS" >&2
+    # libpython3.so carries a bogus $ORIGIN NEEDED entry. The bundled fallback
+    # git wants Debian's libcurl-gnutls ABI; the app prefers git from PATH.
+    # shellcheck disable=SC2016
+    @autoFormatelf@ -j 0 \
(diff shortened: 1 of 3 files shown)


abkein:
My idea was, instead of vendoring the ~1 GB bundle, to drop the common components (python, node, etc) and keep only the proprietary Python and Node modules thereby reducing the runtime size from 1.8 GB to ~760 Mb (excluding shared pkgs in the store). The bundle itself seems to be as proprietary as ChatGPT Desktop app itself.

It's all about the priorities though.

Mic92:
> My idea was, instead of vendoring the ~1 GB bundle, to drop the common components (python, node, etc) and keep only the proprietary Python and Node modules thereby reducing the runtime size from 1.8 GB to ~760 Mb (excluding shared pkgs in the store). The bundle itself seems to be as proprietary as ChatGPT Desktop app itself.
> 
> It's all about the priorities though.

It's just these patches look very brittle: https://github.com/numtide/llm-agents.nix/pull/8340/changes#diff-df647d564ab83d1eb857231cfa207fc472678e75a307c4e390abcf7ce5af8601R31
and using bubblewrap also has sandbox sideeffects. I am not sure what kind of sandboxing they do apply on top of that but it might not compose well.

packages/chatgpt/patch-runtime.sh
```diff
@@ -0,0 +1,41 @@
...
+    printf '%s' "$want" >"$stamp"
+  done
+}
+
+chatgpt_patch_runtimes
+# The runtime may be (re)downloaded while the app runs; patch it once it lands
+# so the first use in this session works too.
+(
+  root="${HOME:-}/.cache/codex-runtimes"
+  for _ in $(seq 120); do
```
Mic92: on the other hand this looks also very brittle... there should be a better way to catch this.

Mic92:
Replaced the poll loop with `inotifywait -e moved_to` on `~/.cache/codex-runtimes`: the installer extracts into a `codex-runtime-install-*` staging dir and `rename()`s it into place, so that event fires exactly once the runtime is complete. The watcher exits within a minute of the app.

PR #8447, opened 2026-08-26
opencode2: follow npm beta channel

StifflyAnyhow:
upstream retired the \`next\` dist-tag as the moving v2 preview channel in august (anomalyco/opencode@9e22c40ff and follow-ups); \`beta\` is the successor, and \`next\` is now frozen at a hand repoint to 0.0.0-beta-17823.

that left opencode2 stuck at 0.0.0-next-17444: every \`0.0.0-beta-*\` sorts below \`0.0.0-next-17444\` in the updater's comparator (equal numeric parts, then prerelease suffix lexically), so the updater declines every beta as an apparent downgrade — a tag flip alone never lands.

- versionSource tag next -> beta
- versionPolicy = "follow_pointer" so channel resets that sort lower aren't skipped. this also wires allow_downgrade through the platform flow, which previously ignored versionPolicy (only manifest-checksums honored it), so the flag isn't a silent no-op here
- hashes.json reseeded to the current beta 0.0.0-beta-18269 via \`nix run .#opencode2.updateScript\`

### DIFF
--- a/packages/opencode2/hashes.json
+++ b/packages/opencode2/hashes.json
@@ -1,8 +1,8 @@
 {
-  "version": "0.0.0-next-17444",
+  "version": "0.0.0-beta-18684",
   "hashes": {
-    "aarch64-darwin": "sha256-kJEZ6dwUayg4qfbGPB1kBX03hqLMpjh/saHcodMRYzs=",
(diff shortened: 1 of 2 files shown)


0xrsydn:
bumping this, i have same issue and the right patch is next -> beta, since current oc2 has a lot of bugs already fixed but this llm-agent.nix version still stalled to the next version one

PR #8743, opened 2026-09-02
clauth: init at 0.15.0

aldoborrero:
Adds [clauth](https://github.com/uwuclxdy/clauth), a Claude Code multi-account manager and usage monitor (CLI, TUI, MCP), built from source with `buildRustPackage`.

The launch-time self-updater is patched out (can't replace a store binary); the three tests asserting `CLAUTH_NO_UPDATE` semantics are skipped accordingly. Shell completions and `versionCheckHook` included.

Tested: `nix build .#clauth` (1776 tests pass), `nix run .#clauth -- --version` → `clauth 0.14.1`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -697,6 +697,16 @@ Nix packages for AI coding agents and development tools. Automatically updated d
 
 </details>
 <details>
+<summary><strong>clauth</strong> - Claude Code multi-account manager and usage monitor (CLI, TUI and MCP cross-account delegation)</summary>
+
+- **Source**: source
+- **License**: MIT
+- **Homepage**: https://github.com/uwuclxdy/clauth
+- **Usage**: `nix run github:numtide/llm-agents.nix#clauth -- --help`
+- **Nix**: [packages/clauth/package.nix](packages/clauth/package.nix)
+
+</details>
+<details>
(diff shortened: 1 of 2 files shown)


zimbatm:
Bumped to 0.15.0 and rebased onto `main`.

The `checks.{x86_64,aarch64}-linux.pkgs-clauth` failure was upstream's own flaky test, `profile::tests::reload_fingerprint_ignores_a_bare_session_token_stamp`: two real writes landing inside one filesystem timestamp tick left the two fingerprints byte-identical at 100 ns precision, so the closing `assert_ne!` read a retired receipt as a live one. Upstream describes it as "1 run in 3 under the full suite" and fixed it in 0.15.0 by backdating the mint an hour, so the re-mint cannot land on the same `SystemTime`. A version bump therefore beats skipping the test.

Checked that the bump does not break the recipe: the `postPatch` anchor in `src/update.rs` is unchanged, all three skipped `updates_enabled_when_env_*` tests still exist in `tests/inline/update.rs`, and the `completions <shell>` subcommand `postInstall` uses is unchanged. `nix build .#clauth` passes locally with the full test suite, `versionCheckHook` reports `clauth 0.15.0`, and `nix fmt`, `ast-grep`, and `meta-completeness` are clean.

Your authorship on the commit is preserved.

PR #8777, opened 2026-09-02
tokscale: init at 4.15.0

smdex:
Add [tokscale](https://github.com/junhoyeo/tokscale) 4.15.0, a CLI and TUI for AI token usage analytics, built from source.

### DIFF
--- a/README.md
+++ b/README.md
@@ -782,6 +782,16 @@ Nix packages for AI coding agents and development tools. Automatically updated d
 - **Nix**: [packages/mindwalk/package.nix](packages/mindwalk/package.nix)
 
 </details>
+<details>
+<summary><strong>tokscale</strong> - CLI and TUI for AI token usage analytics</summary>
+
+- **Source**: source
+- **License**: MIT
+- **Homepage**: https://github.com/junhoyeo/tokscale
+- **Usage**: `nix run github:numtide/llm-agents.nix#tokscale -- --help`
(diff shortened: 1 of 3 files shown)


zimbatm:
Reviewed. The package builds and the recipe is sound; two things before merge.

**`packages/tokscale/default.nix` is dead code.** The flake loads `packages/<name>/package.nix` directly (`flake.nix`: `self.callPackage (./packages + "/${name}/package.nix")`), so this forwarding file is never read. 06b397656 ("packages: drop forwarding-only default.nix files") removed all 107 of them, and no package in the tree has one today. Declaring `flake` as a `package.nix` argument is enough — the scope provides it, which is what your `package.nix` already does. Please drop the file.

**The `postPatch` does more than its comment says.** The comment explains the OpenSSL swap, but the same block also turns off link-time optimisation and raises codegen units:

```nix
--replace-fail 'lto = true' 'lto = false' \
--replace-fail 'codegen-units = 1' 'codegen-units = 16'
```

Those are upstream's deliberate release-profile settings, and disabling them ships a measurably slower binary to every user to save build time once. If that trade is intended, it needs its own comment saying so; otherwise the two lines should go.

Minor, take or leave: `platforms = platforms.all` includes Windows, but the derivation has a darwin-specific `libiconv` branch and nothing for Windows — `platforms.unix` matches what this actually supports. And the other three PRs in this batch use `finalAttrs` rather than `rec`.

PR #8778, opened 2026-09-02
multica, multica-desktop: init at 0.4.38

smdex:
Add [multica](https://github.com/multica-ai/multica) 0.4.38 (Go CLI) and multica-desktop (Electron client built from source against nixpkgs electron, main process deps bundled so no runtime node_modules is shipped). Unfree.

### DIFF
--- a/README.md
+++ b/README.md
@@ -604,6 +604,26 @@ Nix packages for AI coding agents and development tools. Automatically updated d
 
 </details>
 <details>
+<summary><strong>multica</strong> - Command-line interface for the Multica platform</summary>
+
+- **Source**: source
+- **License**: unfree
+- **Homepage**: https://github.com/multica-ai/multica
+- **Usage**: `nix run github:numtide/llm-agents.nix#multica -- --help`
+- **Nix**: [packages/multica/package.nix](packages/multica/package.nix)
+
+</details>
+<details>
+<summary><strong>multica-desktop</strong> - Desktop client for the Multica platform</summary>
+
+- **Source**: source
+- **License**: unfree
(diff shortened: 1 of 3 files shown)


zimbatm:
Reviewed. Nice work on the pruning script — the ELF check and the resolve-verification pass are the right shape for an Electron tree. Three things before merge.

**Both `default.nix` files are dead code.** The flake loads `packages/<name>/package.nix` directly (`flake.nix`: `self.callPackage (./packages + "/${name}/package.nix")`), so neither file is ever read. 06b397656 ("packages: drop forwarding-only default.nix files") removed all 107 of them and none remain in the tree. That includes `multica-desktop/default.nix`, whose `inherit (perSystem.self) multica` is unnecessary: the scope already contains every in-repo package, so declaring `multica` as a `package.nix` argument — which you already do — resolves it. Please drop both.

**Stale comment in `multica/package.nix`:**

```nix
# v0.4.30 pins go 1.26.6 in server/go.mod but nixpkgs go_1_26 is 1.26.5.
```

The package is at 0.4.38. Worth re-checking whether the `unpinGoModVersionHook` is still needed at this version, and updating the version in the comment either way.

**The frozen `ldflags` will rot:**

```nix
"-X main.commit=d563bfbc0"
"-X main.date=2026-08-19T10:03:12Z"
```

`nix-update` bumps `version`, `hash` and `vendorHash`, but nothing will ever touch these two. After the first bump the binary reports a commit and a build date belonging to a different release. Either drop them and let the upstream defaults stand, or derive them from `src`.

Minor: `doCheck = false` has no comment, while the other PRs in this batch each explain why tests are off.

PR #9183, opened 2026-09-11
kandev: 0.91.0 -> 0.94.0

zaninime:
Bump kandev/kandev-desktop to 0.94.0 and rebase `prefer-native-acp-runtimes.patch`.

Second commit decouples `goModules` from the patch (`overrideModAttrs`) so nix-update can prefetch `vendorHash` even when the patch no longer applies — that is why the bot never opened this PR itself — and extends the FHS-PATH test fix to the new goose install-script test.

### DIFF
--- a/packages/kandev/package.nix
+++ b/packages/kandev/package.nix
@@ -54,13 +54,13 @@
 
 let
   pname = "kandev";
-  version = "0.91.0";
+  version = "0.94.0";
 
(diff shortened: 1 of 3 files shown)


packages/kandev/package.nix
mulatta: ` apps/backend/internal/agent/agents/goose_acp_test.go \` should be added line 205

PR #9254, opened 2026-09-13
agent-slack: init at 0.10.2

Ramblurr:
CLI for AI agents to read/send Slack messages, search and manage channels. Built from source with bun, `bun-github` updater.


Ramblurr:
I force pushed a better version that builds from source instead of using their binaries.

PR #9277, opened 2026-09-14
opencode2: follow @opencode/cli stable channel

StifflyAnyhow:
Upstream shipped V2 stable and renamed the npm scope `@opencode-ai/*` → `@opencode/*`. The old scope's `beta` dist-tag is frozen at `0.0.0-beta-19271`, so opencode2 has been tracking a dead channel since the rename.

- `versionSource`/`urlTemplate` repointed to `@opencode/cli(-<platform>)`, following `latest` (the stable 2.x line; the npm source defaults to it)
- tarball binary renamed upstream (`bin/opencode2` → `bin/opencode`); still installed as `opencode2` to coexist with the v1 package
- `hashes.json` reseeded to 2.0.3 via the update script; `meta`/`downloadPage` refreshed

Follow-up to #8447. Whether the v1 `opencode` package should move to the 2.x line is left for a separate change.


Mic92:
Will v1 still receive updates?

PR #9278, opened 2026-09-14
omp: 18.1.19 -> 18.2.0

sernl:
Bump to 18.2.0. opusic-sys now builds bundled libopus via cmake/ninja; collab-cli.ts bare `chalk` import is redirected to the pi-utils re-export.

Based on #9264 (needs bun 1.4.2 from nixpkgs); merge that first. Supersedes #9268.


zimbatm (commented):
Thanks. This is now covered elsewhere: the updater opened #9345 for omp 18.2.1 (with the cmake fix that opusic-sys needs), and the nixpkgs bump is #9264. Once those land this PR can be closed.

PR #9283, opened 2026-09-14
grok-bot: init at 0.47.0

jonjitsu:
## Summary

  Add `grok-bot` 0.47.0: the Grok Bot desktop agent (xAI), packaged from the
  prebuilt Electron `.deb` on Cursor's download CDN for `x86_64-linux` and
  `aarch64-linux`.

  **Package**
  - Build follows `orca`: `.deb` unpacked with `ar`, `formatelf` resolves the
    linked libs, dlopen()ed libs go in `runtimeDependencies`, and `libglvnd` is
    added to the wrapper's `LD_LIBRARY_PATH` so ANGLE finds the native libEGL.
  - `chrome-sandbox` is removed (a store path can't be setuid). `--no-sandbox` is
    added because upstream's Electron build crash-loops sandboxed webview
    renderers (`FATAL:platform_shared_memory_region_posix.cc`); upstream already
    disables the sandbox for most other processes.
  - The wrapper sets `CHROME_DESKTOP=grok-bot.desktop` so Electron registers the
    URL handler against our desktop entry. It also adds the usual
    `NIXOS_OZONE_WL` Wayland flags.
  - The desktop entry mirrors upstream's, including the `x-scheme-handler/grokbot`
    and `x-scheme-handler/sand` login-redirect handlers.
  - No `versionCheckHook`: GUI-only app, `--version` would start Electron.

  **Updater**
  - Custom `update.py` rather than a declarative `passthru.updater`: the
    download URLs embed a 40-char build id
    (`.../grokbot/stable/<buildId>/linux/<arch>/grok-bot_<version>_<arch>.deb`)
    that can't be derived from the version, so the `platform` kind can't
    template them.
  - It queries Cursor's stable update feed for the version and build id,
    prefetches both `.deb`s, and writes `urls`/`hashes` to `hashes.json`. It never
    downgrades, even if a staged rollout serves an older build.

  **Other**
  - Adds `jonjitsu` to the custom maintainers in `lib/default.nix`.
  - README package list regenerated.

  ## Test plan

  - [x] `nix build .#grok-bot` succeeds
  - [x] `nix run .#grok-bot` launches the app (login not yet tested)
  - [x] Package has a custom `update.py` since nix-update can't handle the
        build-id URLs:
    - run against the live feed: `grok-bot: already up to date (0.47.0, feed: 0.47.0)`
    - with the version set back to 0.46.0 it prefetched both `.deb`s and
      restored an identical `hashes.json`
  - [x] `ruff check`, `ruff format --check`, `mypy`, `nixfmt --check`,
        `ast-grep scan` pass on the new files
  - [x] `nix fmt` pass
  - [] `nix flake check` pass

Used a combination of Cursor and Claude code to generate this.


zimbatm (approved):
Reviewed: binary Electron .deb from downloads.cursor.com (matches the live update feed), unfree + binaryNativeCode declared, maintainer id verified, updater justified, nixbot green on all systems, local build output checked.

Issue #6573, opened 2026-07-08
package request: Snowflake Cortex Code CLI


frankzvitale:
### Tool name

Cortex Code CLI

### Project URL

https://docs.snowflake.com/en/user-guide/cortex-code/cortex-code-cli

### Will you package and maintain this?

Yes, I'll submit a PR and maintain it

### Have you tested this tool?

Yes, I use it regularly

### Why is this tool worth packaging?

This is Snowflake's native, in-house cli tool for interfacing with warehouses, data engineering pipelines, and analytics workloads. I use this on a daily basis as a data engineer; it is incredibly helpful and under fairly rapid and active development by the Snowflake team, judging by how often updates are pushed.

It ships with Snowflake-specific tooling to enable the agent to interface with Snowflake objects and entities using natural language from the user, and I believe it uses one of Snowflake's text-to-SQL models for actually generating the queries that it runs. This makes it much "smarter" for running against Snowflake data since it already has knowledge and context shipped by the Snowflake team.

Furthermore, and perhaps most crucially, by default inference runs on Snowflake compute-hosted models including frontier OpenAI and Anthropic models, though many open source models are also included as well (cloud-provider and region-dependent, e.g. different Azure regions may have different models available). This is huge for governance as it means that all inference runs entirely within Snowflake infrastructure and that data is governed by a company's preexisting data contract with Snowflake, so it can be used in sensitive environments out of the box where something like Claude Code or Codex may require more configuration to have a similar governance policy.

Finally, because the models it uses run within Snowflake compute, this also allows orgs to maintain a centralized billing model where they can administer model access and rate limits directly from their Snowflake account, without having to provision different business/enterprise accounts with Anthropic, OpenAI, etc. All compute is billed as credits on Snowflake warehouses at their billing rates (accessible online, not particularly relevant here). This makes it a compelling option for some businesses who may want one monolithic structure rather than fraction their spend across different services.

I love this tool and use it constantly, and will be packaging it for my own personal usage on Nix regardless of whether it's added here, but I would rather see it part of this incredible community resource so others can benefit from it as well! Many thanks to the team running this project.

### Current installation method

curl -LsS https://ai.snowflake.com/static/cc-scripts/install.sh | sh

Mic92:
[...]

PR #9329, opened 2026-09-15
coco: init at 1.1.87

frankzvitale:
## Summary

Add Snowflake's CoCo (Cortex Code) CLI at `1.1.87+175514.04f5c9114e11` for x86_64 Linux, aarch64 Linux, and Apple Silicon. The installed command is `cortex`.

The package preserves the Bun payload, supplies runtime tools and Linux seccomp assets, and stages the bundled OpenSSL configuration in a per-user cache when upstream includes it. ARM Linux omits that bundle and starts without it. On macOS, the package replaces the Homebrew-linked ripgrep and re-signs the patched executable.

Two guarded, length-preserving edits let the release-path parser recognize the Nix store installation and default automatic updates to off. Unexpected patch matches fail the build. This follows the repository's existing approach to patching distributed artifacts, including the embedded JavaScript patches in `packages/chatgpt/patch-asar.py`.

The custom updater follows Snowflake's stable pointer, validates the release manifest, and updates all three hashes. It remains custom because the shared URL interpolator does not encode the `+` in Snowflake's version paths; unencoded requests return 404. It also follows upstream rollbacks.

Includes package metadata, maintainer registration, generated README documentation, and package install checks. Native ARM CI is maintained separately on the fork and is not included in this PR. Its jobs build CoCo locally on each runner while allowing cached dependencies, ensuring the install checks execute.

## Test plan

- [x] `nix build --accept-flake-config .#coco` succeeds on x86_64 Linux.
- [x] `nix flake check --accept-flake-config` passes on x86_64 Linux.
- [x] `nix flake check --accept-flake-config --all-systems --no-build` passes.
- [x] Native aarch64 Linux and aarch64 Darwin builds, install checks, and CLI smoke tests pass at pre-squash commit `8ff89861b`, whose package files are identical to submission commit `96273686f`: https://github.com/frankzvitale/llm-agents.nix/actions/runs/34990542372
- [x] `nix fmt`, README generation, and the new-package maintainer check pass. The fork-only CI workflow also passes actionlint.
- [x] Custom updater tested against the live stable manifest with temporarily older, newer, and unchanged versions; the resulting three hashes match the committed pin.

Install checks cover concurrent first launches into one fresh cache, exact version output, connection/MCP listing, shell completions, default update behavior, runtime tools, and FIPS initialization where bundled. Linux also exercises the seccomp helper. A separate local control check confirmed an allowed command succeeds and AF_UNIX socket creation fails with EPERM under the filter.

Sample output from a fresh HOME:

```text
$ cortex --version
Cortex Code v1.1.87
$ cortex connections list
{
  "active_connection": null,
  "connections": {}
}
$ cortex mcp list
No MCP servers configured.
```

Manual testing on NixOS under WSL also exercised Snowflake object search and project tooling; the tester reported correct search results and that the other tooling worked well. Automated packaging checks use no account credentials.

One unresolved observation from the WSL environment: dbt emits a `RequestsDependencyWarning`, without an observed functional failure. The warning was not reproduced by importing the bundled Snowflake connector in a clean environment; the dbt interpreter and dependency source have not been established. The tester considers this non-blocking.

The bundled Python REPL is independent of a project virtualenv by default; upstream provides `COCO_PYTHON_REPL_PYTHON_PATH` to select a local interpreter. FIPS tests verify runtime initialization, not certification.


frankzvitale:
@Mic92 good afternoon! Delayed more than I'd have liked but here's my contribution following up on #6573 - please let me know how this looks and whether there's anything else I need to do here! Thanks so much.

frankzvitale:
Something I may look into for a further improvement to reduce the code in here is potentially expanding the manifest-checksums updater to handle URL encoding; Snowflake uses literal + symbols in their versions which breaks the current literal {version} substituter in the manifest-checksums updater, which gives 404 errors when the manifest URL is queried.

Will probably make this a follow up issue for discussion, but for now this comes with the custom updater.

zimbatm (commented):
CI is red on `treefmt` (all three systems). ruff rejects `packages/coco/patch-payload.py`:

- `INP001`: add `# ruff: noqa: INP001` as the first line, like `packages/plannotator/fix-bun-offline-install.py`.
- `S101`: replace the `assert` with an explicit check, e.g. `if data.count(old) != 1: raise ValueError("auto-update option not found exactly once")`.

Log: https://nixbot.numtide.com/repos/github/numtide/llm-agents.nix/builds/5915/logs/checks.x86_64-linux.treefmt

Also: this PR changes `lib/`, `flake.nix`, `checks/`, and `scripts/updater/` alongside the new package. I will leave the merge call on those shared changes to @zimbatm.

frankzvitale:
@Mic92 @zimbatm I am taking a look at these changes and will look at the CI failures as well - thanks!

Issue #9364, opened 2026-09-16
cursor-agent: installing into home.packages/buildEnv conflicts on $out/index.js (e.g. with playwright)


9bingyin:
### Package

cursor-agent

### Issue type

Other

### Platform

aarch64-darwin

### Nix sandbox enabled?

No

### Description

`cursor-agent` copies the upstream `dist-package` contents to `$out/` (`cp -r dist-package/* $out/`), so files like `index.js`, `node`, and `rg` live at the package root.

That layout cannot be merged into a shared profile via `pkgs.buildEnv` / home-manager's `home.packages`: any other package that also ships `$out/index.js` (for example nixpkgs `playwright`, which is currently `playwright-core`) causes a conflicting-subpath failure.

**Minimal reproduction** (aarch64-darwin):

```nix
# flake inputs: llm-agents.nix + nixos-unstable
pkgs.buildEnv {
  name = "repro-cursor-agent-conflict";
  paths = [
    inputs.llm-agents.packages.${system}.cursor-agent
    pkgs.playwright # → playwright-core-1.61.1
  ];
}
```

Observed with `cursor-agent-2026.09.10-fd3934a`.

**Workaround:** wrap so only `/bin` is linked into the profile:

```nix
pkgs.buildEnv {
  name = "cursor-agent-bin";
  paths = [ inputs.llm-agents.packages.${system}.cursor-agent ];
  pathsToLink = [ "/bin" ];
}
```

**Suggested fix:** install like nixpkgs `cursor-cli` — put the dist under `$out/share/cursor-agent/` and expose `$out/bin/cursor-agent` (symlink or wrapper). Upstream `cursor-agent` already resolves siblings via `realpath`/`SCRIPT_DIR`, and nixpkgs uses that layout successfully:

https://github.com/NixOS/nixpkgs/blob/nixos-unstable/pkgs/by-name/cu/cursor-cli/package.nix

That would make `home.packages = [ cursor-agent … ]` safe without a local `pathsToLink` workaround.

### Error output / logs

```text
pkgs.buildEnv error: two given paths contain a conflicting subpath:
  `/nix/store/bm9fxdc3qy2154z7kk4cpnj0wqj94ylh-playwright-core-1.61.1/index.js' and
  `/nix/store/fladdbdl1wl7py08g4a91iw6ha4fdaa5-cursor-agent-2026.09.10-fd3934a/index.js'
[...]

PR #9376, opened 2026-09-16
Enforce an FHS-like root for every package output

zimbatm:
## Summary

Follow-up to #9364 / #9375, which was one instance of a general problem: whatever a package puts directly under `$out` lands in the top level of every `pkgs.buildEnv` and home-manager profile, and two packages with the same root file name collide.

A survey of every `packages.<system>` output (197 on x86_64-linux, 184 on aarch64-darwin, root listings read from the binary cache) found five offenders. Each gets its own commit:

| package | before | after |
| --- | --- | --- |
| cursor-agent | npm dist at the root | `share/cursor-agent/` (same as #9375, cherry-picked) |
| hunk | `skills/` at the root | `share/hunk/{bin,skills}`; the binary walks up from `process.execPath` for `skills/`, so a makeWrapper wrapper keeps that path real |
| site | document root at `$out` | `share/llm-agents-site/`; `lib/hercules-ci.nix` copies that subdirectory to gh-pages |
| flake-inputs | output was a regular file | `nix-support/flake-inputs` |
| codex (Linux) | `codex-resources/` at the root | `libexec/codex/{bin,codex-resources}`; codex looks for a bundled bwrap next to its own executable and for `codex-code-mode-host` in its own directory, so binaries and resources move together |

`checks/fhs-layout.nix` enforces the rule from now on: one derivation that depends on every package output, lists each root, and fails with the package name and the stray entries when anything outside `bin sbin lib lib64 libexec share include etc opt Applications nix-support` appears. It also fails on an output that is not a directory, and cross-checks the number of roots inspected against the number of packages. The `exceptions` set is empty.

AGENTS.md gains an "Output Layout" section with the allowlist and the `share/<pname>` convention.

## Test plan

- [x] the five changed packages build on x86_64-linux and answer `--version` (codex-cli 0.154.0, hunk 0.22.0, cursor-agent 2026.09.10-fd3934a); `hunk skill path hunk-review` resolves under the new tree
- [x] `nix build .#checks.x86_64-linux.fhs-layout`: "All package roots are FHS-like" (builds all 197 packages)
- [x] the check was pointed at a stray root on purpose and failed as expected
- [x] `nix fmt`, `ast-grep scan packages`, `nix flake check --no-build --all-systems`
- [ ] aarch64-darwin builds (nixbot); hunk's `process.execPath` lookup is the one darwin-specific risk


AGENTS.md
```diff
@@ -152,6 +152,45 @@ stdenv.mkDerivation rec {
 }
 ```
 
+### Output Layout
+
+A package output must have an FHS-like root. The only entries allowed directly
```
Mic92, line 157: can we shorten this?

packages/hunk/package.nix
```diff
@@ -70,13 +72,17 @@ stdenv.mkDerivation {
     runHook postBuild
   '';
 
+  # The binary locates its bundled skills by walking the ancestors of
+  # process.execPath for skills/<name>/SKILL.md (packages/hunk/src/core/run/
+  # paths.ts), so the skills have to sit above the real executable. Both live
+  # under share/hunk and $out/bin/hunk is a wrapper, which keeps the package
+  # root free of a generic `skills` directory that collided with other
+  # packages in buildEnv and home-manager profiles (#9364).
```
Mic92, line 80: too long.

packages/flake-inputs/package.nix
```diff
@@ -2,9 +2,13 @@
   inputs,
   pkgs,
 }:
-# A derivation that references all flake inputs to ensure they get cached
+# A derivation that references all flake inputs to ensure they get cached.
+# The reference list goes under nix-support/ so the output is an FHS-like
+# directory rather than a bare file (see checks/fhs-layout.nix).
```
Mic92, line 7: why do we care about this one? I think a file is fine here.

packages/codex/package.nix
```diff
@@ -89,12 +89,23 @@ rustPlatform.buildRustPackage (
 
     inherit preBuild;
 
+    # codex accepts a bundled bwrap only at <exe dir>/codex-resources/bwrap or
```
Mic92, line 92: too much text.

checks/fhs-layout.nix
```diff
@@ -0,0 +1,104 @@
...
+    "lib"
+    "lib64"
+    "libexec"
+    "nix-support"
+    "opt"
+    "sbin"
+    "share"
+  ];
+
+  # Packages whose launcher hardcodes its own root and that therefore cannot
```
Mic92, line 38: too much text.

PR #9384, opened 2026-09-16
feat: add yylo at 0.2.3

InsightFactoryAPP:
Adds **yylo** (`@yylo/cli` on npm) — task-driven AI subagent orchestration for coding agents (MIT, node >= 20.10). Mirrors the `packages/skills` pattern: npm-registry tarball (`hashes.json` source hash) + vendored `package-lock.json` injected via `srcWithLock`, `npmDepsFetcherVersion = 2`, declarative `mkUpdater` (`kind = "npm"`, purl `pkg:npm/@yylo/cli`).

Two integration notes specific to this package:

- The published `package.json` carries repo-only lifecycle scripts (`prepack` rebuilds from `src/`, which the tarball does not ship). `srcWithLock` strips `scripts` with `jq` so `npm pack`/`prune` operate on the shipped, already-built `dist/`.
- The `yylo`/`yy`/`ypl` bins are bash wrappers that exec the real node CLI, so they are re-wrapped with `makeWrapper` (`nodejs`, `bash`, `coreutils`, `gnugrep`, `gnused` on PATH); `feedback-yylo` is a plain `.mjs` and keeps its default shim. The wrapper's optional project-bootstrap path (only inside an initialized project) expects `python3` from the user environment — kept out of the closure deliberately.

Verified: `nix build --accept-flake-config .#yylo` green incl. `versionCheckHook` (`yylo --version` → `0.2.3`), `nix fmt`, `./scripts/generate-package-docs.nu`, `nix flake check --no-build`, and the `meta-maintainers` / `meta-completeness` / `treefmt` / `updater-tests` / `interpolate-conformance` checks all pass.


Mic92:
This package needs more work.

PR #9429, opened 2026-09-17
omp: 18.2.0 -> 18.2.4

arch-fan:
## Summary

🤖 **AI-generated PR notice:** This PR was created by an AI agent (Hermes Agent by Nous Research) running in @arch-fan's environment, on his behalf. The change was developed, built, and tested by that agent; @arch-fan takes responsibility for the submission.

Bump `omp` from 18.2.0 to 18.2.4 on top of current `main`.

- `main` already contains 18.2.0 (via the merge of #9278), including the `opusic-sys` cmake/ninja fixups, so this PR reduces to the `bun.nix` / `hashes.json` refresh only.
- The `hashes.json` / `bun.nix` values are those generated by the updater on #9345 (same content), pre-rebased onto current `main` so the branch applies without conflicts.
- Background for the fixups that made 18.2.x work: omp 18.2.x compiles its standalone binary with `bytecode: true`, and bytecode emitted by bun 1.3.x fails to load the bundle under the 1.3.x runtime (`TypeError: Expected CommonJS module to have a function wrapper`). With nixpkgs' bun 1.4.2 (now on `main`) the binary loads and the smoke test passes.

## Test plan

- [x] Package updates via `nix-update` or has a custom `update.py` if nix-update doesn't work — `omp` has a custom `update.py`; the values are the updater's (from #9345).

On nixbot, these exact 18.2.4 values were already built in #4693: `checks.x86_64-linux.pkgs-omp` and `checks.aarch64-darwin.pkgs-omp` pass. The only red attribute there is `checks.aarch64-linux.pkgs-omp`, which fails while the build farm distributes `bun-pkg` fixed-output derivations to the aarch64 builders (`gRPC AddMultipleToStore ... error: store path ... does not exist`, alongside `worker shutting down, build interrupted`) — an infra issue that has hit several builds today, including #9345 itself and an earlier iteration of this PR.

Local verification of the identical packaging approach at 18.2.3 (x86_64-linux, full cargo build, Nix container):

```
omp> buildPhase completed in 21 minutes 35 seconds
omp> Running phase: installCheckPhase
$ ./result/bin/omp --version
omp/18.2.3
$ HOME=$TMPDIR ./result/bin/omp --smoke-test
smoke-test: ok
$ BUN_BE_BUN=1 ./result/lib/omp/omp -e 'console.log(Bun.version, typeof Bun.Image)'
1.4.2 function
```

## Related

- #9345 — updater PR (now 18.2.4); this is its content rebased cleanly onto current `main` (maintainers can take whichever path they prefer)
- #9278 — 18.2.0, merged


arch-fan:
Superseded by #9441 — branch renamed to `omp-18.2.4` for accuracy; content unchanged. *(AI-generated note on behalf of @arch-fan.)*

PR #9477, opened 2026-09-18
prime-agent: 0.9.3 -> 0.9.5

xenia-foxtrot:
## Summary

 - Update Prime Agent from 0.9.3 to 0.9.5.
 - Remove the obsolete model-generation substitution that broke automatic updates.
 - Update Python runtime checks for the new APIs.
 - Scope MCP 2 and its dependencies to Prime Agent’s Python environment

## Test plan

- Package build and install checks passed
- Formatting, metadata checks, and `nix flake check` passed.
- 10 external MCP runtime tests passed, covering stdio, HTTP/SSE, and timeouts.


mulatta:
I think overriding mcp version 2 make maintain harder. If we override it, we should maintain too many dependencies. 
also, still prime-agent is work well with `pythonRelaxDeps` because prime-agent 0.9.3..0.9.5 [mcp_base.py](https://github.com/PrimeIntellect-ai/prime-agent/blob/a7d791bc1be09793ed5f3ec05bf4cccbc60679ea/prime-agent-runtime/src/rlm/mcp_base.py) only changed its inputSchema 

Also, nixpkgs mcp2 PR looks breaking changes for other packages. [nixpkgs#556839](https://github.com/NixOS/nixpkgs/pull/556839)

packages/prime-agent/package.nix
```diff
@@ -51,11 +52,6 @@ buildNpmPackage (finalAttrs: {
   postPatch = ''
     cp ${./package-lock.json} package-lock.json
 
-    # Release tags include generated model data. Regenerating it would access
```
mulatta: We need this patch removal 👍

packages/prime-agent/package.nix
```diff
@@ -89,150 +85,356 @@ buildNpmPackage (finalAttrs: {
...
-      dependencies = with python3.pkgs; [
-        ipykernel
-        mcp
-        nest-asyncio
-        tyro
-      ];
-      # Pins mcp>=2 but rlm/mcp_base.py explicitly supports the 1.x client API.
-      pythonRelaxDeps = [ "mcp" ];
-    };
+  passthru =
```
mulatta: Python overrides are not required when we use `pythonRelaxDeps = [mcp];`

packages/prime-agent/package.nix
```diff
@@ -89,150 +85,356 @@ buildNpmPackage (finalAttrs: {
...
-        requests
-        scipy
-        tomli
-        tyro
-      ])
-      ++ [ finalAttrs.passthru.primeAgentRuntime ]
-      ++ finalAttrs.passthru.pythonSkills
-    );
-  };
+            uv-dynamic-versioning = pyPrev.uv-dynamic-versioning.overridePythonAttrs (old: {
```
mulatta: All those overrides are not required.

packages/prime-agent/package.nix
```diff
@@ -89,150 +85,356 @@ buildNpmPackage (finalAttrs: {
   ];
 
   postInstallCheck = ''
```
mulatta: I think only change `assert callable(rlm.run)` -> `assert callable(rlm.spawn)` is enough

xenia-foxtrot:
There's actually a single call site in prime-agent where MCP 2 and MCP 1's API differ enough to cause an error. I have been using a patch for prime-agent to allow for MCP 1, I can push a new change to include that patch if you'd like. 

I opted to go for MCP 2 here, as it's what prime-agent expects, and we can scope it to the prime-agent environment.

mulatta:
#9627 has been merged. 
I think if the patch is small and maintainable, it is better..

Wizermil:
there is definitely a MCP issue with the current version.
I hacked my nix config to test the PR. I confirm that with this PR MCPs work way better out of the box. No need for monkey patching prime-agent
```
    prime-agent-mcp-fix = {
      url = "github:xenia-foxtrot/llm-agents.nix/e6f7d25119f391b696e0297f71e81e2173cb8fa2";
      flake = false;
    };
```
and
```
    prime-agent = pkgs.callPackage (inputs.prime-agent-mcp-fix + "/packages/prime-agent/package.nix") {
      flake = inputs.llm-agents;
      versionCheckHomeHook = agentsRaw.versionCheckHomeHook;
    };
```

xenia-foxtrot:
Thanks @Wizermil for testing!

#9627 covers the version bump and build fixes, but the MCP runtime issue remains. The incompatible call is in `prime-agent-runtime/src/rlm/mcp.py`, rather than `mcp_base.py`.

Prime Agent passes a numeric value to `ClientSession` for the timeout, as is the API in MCP 2. MCP 1 expects a `datetime.timedelta` and calls `.total_seconds()` on it, which results in:

`AttributeError: 'float' object has no attribute 'total_seconds'`

pythonRelaxDeps bypasses the dependency-version check, but doesn’t resolve that API difference.

I understand the maintenance concern, but I'd still prefer using MCP 2, since that's the version that Prime Agent explicitly requires. This failure demonstrates that relaxing the requirement alone is insufficient; patching that call fixes the known mismatch but still leaves us responsible for validating MCP 1. 

The overrides are scoped to Prime Agent's Python environment, so other Nixpkgs consumers keep their existing MCP version. That addresses the cross-package breakage concern, although I agree the private package definitions add upkeep.

xenia-foxtrot:
I've discovered that we can remove the private definitions for some of the dependencies; the ones in `nixpkgs` are new enough. I've updated the PR so we only privately package `MCP 2.0.0` and `mcp-types 2.0.0`.

Once MCP 2.0 lands in nixpkgs, we can remove the private packaging. 

Does this ease maintenance concerns?

mulatta:
Thanks for simplifying this. 
Limiting the override to just mcp and mcp-types addresses my maintenance concerns.

zimbatm (approved):
MCP 2 stays inside a prime-agent-private python3 package set, so nothing else in the tree rebuilds against it. mcp and mcp-types come from the modelcontextprotocol/python-sdk v2.0.0 tag, and the postInstallCheck now pins the installed versions and the skill list. Note the description still claims a 0.9.3 to 0.9.5 bump, which this diff does not contain.

